PR #3, opened 2019-12-17
docs: create initial structure

mvidalgarcia:
closes #1

PR #5, opened 2019-12-18
readme: add running steps

mvidalgarcia:
Addresses https://github.com/reanahub/docs.reana.io/pull/3#discussion_r359218831

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,16 @@
 # docs.reana.io
-REANA user documentation web site
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,2 +1,11 @@
 # docs.reana.io
 REANA user documentation web site
```
tiborsimko: Please add `## About` here, also readme and license badges, so that we have a nice harmony with `www.reana.io`

mvidalgarcia: Updated.

Issue #36, opened 2020-03-31
docs: vomsproxy auth


diegodelemos:
Document VOMS proxy usage in http://docs.reana.io/advanced-usage/access-control/vomsproxy/

```console
$ read -s VOMSPROXY_PASS
$ reana-client secrets-add --file userkey.pem \
                           --file usercert.pem \
                           --env VOMSPROXY_VO=cms \ 
                           --env VOMSPROXY_PASS=$VOMSPROXY_PASS
```

Implemented in https://github.com/reanahub/reana/issues/256.

PR #37, opened 2020-04-01
voms-proxy: add docs for using voms_proxy field

alintulu:
Currently only for serial. Not sure if field `Testing your credentials` is necessary?

closes  reanahub/docs.reana.io#36

### DIFF
--- a/docs/advanced-usage/access-control/voms-proxy/index.md
+++ b/docs/advanced-usage/access-control/voms-proxy/index.md
@@ -1,0 +1,68 @@
+# VOMS-proxy
+
+If your workflow needs special permissions that your Virtual Organization (VO) entitles you to, you can use VOMS  authentication. VOMS (Virtual Organization Membership Service) provides information on the user's relationship with their VO. VOs administer users and grant them permissions according to their group, role and capability. At CERN the different experiments such as ALICE, ATLAS, CMS and LHCb all have their own VO. Being part of a VO can give you access to globally distributed data analysis infrastructure.
+
+If you do not already have a user certificate create one at [ca.cern.ch/ca](https://ca.cern.ch/ca/).
+
+## Testing your credentials
+
+Before moving to REANA, check that your credentials are in order.
+
+```console
+# Login to lxplus
+$ ssh johndoe@lxplus.cern.ch
+
+# Make sure userkey.pem is read/write only by the owner
+$ ls -l ~/.globus
+-rw-r--r--. usercert.pem
+-r--------. userkey.pem
+
+# Let us create a proxy certificate with the VO cms
+$ voms-proxy-init --voms cms
+Enter GRID pass phrase for this identity:
+
+Contacting voms2.cern.ch:15002 [/DC=ch/DC=cern/OU=computers/CN=voms2.cern.ch] "cms"...
+Remote VOMS server contacted succesfully.
+
+
+Created proxy in /tmp/x509up_u131816.
+
+Your proxy is valid until Wed Apr 01 00:43:08 CEST 2020
+```
+
+## Uploading secrets
+
+To create the proxy certificate REANA needs access to your user certificate `usercert.pem`, encrypted private key `userkey.pem` and GRID pass phrase encoded in Base64. To encode your pass phrase exchange `mygridpassphrase` for your GRID pass phrase and perform:
+
+```console
+$ echo 'mygridpassphrase' | base64
+bXlncmlkcGFzc3BocmFzZQo=
+```
+
+Upload these three secrets to REANA. In addition REANA needs information about your VO. This is communicated by the environment variable `VOMSPROXY_VO` and can also be added to the secrets:
+
+```console
+$ reana-client secrets-add --file userkey.pem \
+                           --file usercert.pem \
+                           --env VOMSPROXY_PASS=bXlncmlkcGFzc3BocmFzZQo= \
+                           --env VOMSPROXY_VO=cms
+```
+
+## Setting voms-proxy requirement
+
+Set `voms_proxy: true` for the steps that need a proxy certificate in the workflow specification.
+Please note that step's docker image (e.g ``environment: 'reanahub/reana-auth-vomsproxy'``)
+should have the VOMS client installed for the command `voms-proxy-info` to work.
+
+Serial example:
+
+```yaml hl_lines="9"
+    workflow:
+      type: serial
+      specification:
+        steps:
+          - environment: 'reanahub/reana-auth-vomsproxy'
+            voms_proxy: true
+            commands:
+            - voms-proxy-info
+```
--- a/docs/advanced-usage/access-control/voms-proxy/index.md
+++ b/docs/advanced-usage/access-control/voms-proxy/index.md
@@ -39,13 +39,13 @@ $ echo 'mygridpassphrase' | base64
 bXlncmlkcGFzc3BocmFzZQo=
 ```
 
-Upload these three secrets to REANA. In addition REANA needs information about your VO. This is communicated by the environment variable `VOMSPROXY_VO` and can also be added to the secrets:
+Upload these three secrets to REANA. In addition REANA needs information about your VO. This is communicated by the environment variable `VONAME` and can also be added to the secrets:
 
 ```console
 $ reana-client secrets-add --file userkey.pem \
                            --file usercert.pem \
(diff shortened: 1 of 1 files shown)


docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,65 @@
+# Voms-proxy
```
diegodelemos, line 1: VOMS in capitals, right?

docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+If your workflow needs special permissions that your Virtual Organization entitles you to, the VOMS client can create a proxy certificate for your authentication.
```
diegodelemos, line 3: What about simplifying the last part?

... your Virtual Organization entitles you to, you can use VOMS authentication.

docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+Created proxy in /tmp/x509up_u131816.
+
+Your proxy is valid until Wed Apr 01 00:43:08 CEST 2020
+```
+
+## Uploading secrets
+
+To create the proxy certificate the VOMS client needs access to your user certificate `usercert.pem`, encrypted private key `userkey.pem` and GRID pass phrase. For REANA the pass phrase needs to be encoded in Base64. To encode it exchange `mygridpassphrase` for your GRID pass phrase and perform
```
diegodelemos, line 32: I don't know if we should rephrase the part that says "the VOMS client needs access to your user certificate" to "REANA needs access to your user certificate", what do you think? For me sounds a bit too much to give details about how it is implemented in the docs. However, I kind of understand it in this case, the users would give their credentials and it is nice to know what part of the system is using it (they are familiar with VOMS client nomenclature).

docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+```console
+$ echo 'mygridpassphrase' | base64
+bXlncmlkcGFzc3BocmFzZQo=
+```
+
+Upload these three secrets to REANA. In addition the client needs information about your virtual organization. This is communicated by the environment variable `VOMSPROXY_VO` and can also be added to the secrets:
```
diegodelemos, line 39: Same here, "the client needs" it might even sound confusing... I would say REANA needs in general, as we do the orchestration of the integration with the VOMS client.

docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,65 @@
...
+If your workflow needs special permissions that your Virtual Organization entitles you to, the VOMS client can create a proxy certificate for your authentication.
+
+## Testing your credentials
+
+Before moving to REANA, check that your credentials are in order.
+
+```console
+# Login to lxplus
+$ ssh johndoe@lxplus.cern.ch
+# Make sure userkey.pem is read/write only by the owner
```
mvidalgarcia, line 12: What about a separation empty line here as you have below? It makes the snippet more readable.

docs/advanced-usage/access-control/voms-proxy/index.md
mvidalgarcia, line 3: I wonder if we should explain better what a Virtual Organization is (I personally don't know 🤷‍♂ ) or perhaps link it to some website that explains the concept? Same for VOMS, what does it stand for and so on.

docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,65 @@
...
+                           --file usercert.pem \
+                           --env VOMSPROXY_PASS=bXlncmlkcGFzc3BocmFzZQo= \
+                           --env VOMSPROXY_VO=cms
+```
+
+## Setting voms-proxy requirement
+
+Set `voms_proxy: true` for the steps that need a proxy certificate in the workflow specification.
+Please note that step's docker image (e.g ``environment: 'reanahub/reana-auth-vomsproxy'``)
+should have the VOMS client installed for the command `voms-proxy-info` to work.
```
diegodelemos, line 52: Here we say that they should have the VOMS client installed in their image, which makes sense... if I would have started reading from here I wouldn't have made any of the previous comments. 

Why my confusion? I think that if they follow these docs they should take for granted that everything is going to work out of the box without having to bother about the details, that's why I would always make reference to REANA as a whole. Although that would be ideal, we have to tell them that the VOMS client has to be installed... so they have to think of how it works, so it is not really "out of the box" (same happens with Kerberos), so all the explanations from before would make sense.

docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,65 @@
...
+```console
+# Login to lxplus
+$ ssh johndoe@lxplus.cern.ch
+# Make sure userkey.pem is read/write only by the owner
+$ ls -l ~/.globus
+-rw-r--r--. usercert.pem
+-r--------. userkey.pem
+
+# Let us create a proxy certificate with the VO cms
+$ voms-proxy-init --voms cms
```
mvidalgarcia, line 18: Should we take for granted that the user is going to have this tool installed? Perhaps we should explain it in the introduction.

docs/advanced-usage/access-control/voms-proxy/index.md
mvidalgarcia, line 32:
```suggestion
To create the proxy certificate the VOMS client needs access to your user certificate `usercert.pem`, encrypted private key `userkey.pem` and GRID pass phrase. For REANA the pass phrase needs to be encoded in Base64. To encode it exchange `mygridpassphrase` for your GRID pass phrase and perform:
```

alintulu: You are right, it does sound better to only mention REANA. I think my thought process was that people are used to the VOMS client and I need to specify usually the client needs this but for REANA we need the pass phrase in Base64. But I agree with you :)

docs/advanced-usage/access-control/voms-proxy/index.md
alintulu: I will remove the mentions of the VOMS client

docs/advanced-usage/access-control/voms-proxy/index.md
alintulu: That's a good idea, I can add some info on VO, something like..(?)

> VOs (Virtual Organizations) administer users and grant them permissions according to their group, role and capability. At CERN the different experiments such as ATLAS, CMS and LHCb all have their own VO. Being part of a VO can give you access to globally distributed data analysis infrastructure and databases. VOMS (Virtual Organization Membership Service) provide information on the user's relationship with their VO.

docs/advanced-usage/access-control/voms-proxy/index.md
alintulu: The VOMS client is available on lxplus however the user does need to download the userkey.pem and usercert.pem themselves! I can explain it in the introduction or I can link to a page like this https://ca.cern.ch/ca/Help/?kbid=024010

docs/advanced-usage/access-control/voms-proxy/index.md
mvidalgarcia: It looks good to me! let's see what @diegodelemos @tiborsimko think

docs/advanced-usage/access-control/voms-proxy/index.md
```diff
@@ -0,0 +1,68 @@
+# VOMS-proxy
+
+If your workflow needs special permissions that your Virtual Organization (VO) entitles you to, you can use VOMS  authentication. VOMS (Virtual Organization Membership Service) provide information on the user's relationship with their VO. VOs administer users and grant them permissions according to their group, role and capability. At CERN the different experiments such as ALICE, ATLAS, CMS and LHCb all have their own VO. Being part of a VO can give you access to globally distributed data analysis infrastructure.
```
mvidalgarcia, line 3: > VOMS (Virtual Organization Membership Service) provide**s**...

Missing "s".

alintulu: fixed!

PR #61, opened 2020-09-22
docs: remove releasing docs

### LINKED ISSUE #60: docs: update release instructions
The release instructions are very incomplete and should be updated.

```
This is incorrect, the `releasehelm` mode is installing  REANA from released Helm charts.

Also, `run-ci` would build Docker images with included shared modules, one has to use `git-submodule --delete` and `docker-build` commands.

I think we need to revamp the releasing section quite a bit.

For this PR, perhaps just leave visible FIXME note, and we shall update it separately.
```

_Originally posted by @tiborsimko in https://github.com/reanahub/docs.reana.io/pull/57#discussion_r483524253_

diegodelemos:
* Closes #60.

Depends on https://github.com/reanahub/reana/pull/361.


docs/development/releasing/index.md
```diff
@@ -2,41 +2,16 @@
 
 ## Helm Charts
 
-### 1. Build and test locally:
+**1.** Upgrade to the latest cluster component versions:
```
diegodelemos, line 5: I thought of keeping it simple, one topic per section. We could add other ones e.g. `## Docker images`, `## Python packages` etc... what do you think?

tiborsimko: Yeah, we'd need to add all the other section if we want to have the full release instruction picture... The chain is quite complex.

diegodelemos: I've included the simplest steps for every category. Ideally, we should migrate https://digital-repositories.web.cern.ch/digital-repositories/reana/releasing/ here. After having a look to it, I have the following comments:
- Shall we move [what is a REANA release](https://digital-repositories.web.cern.ch/digital-repositories/reana/releasing/#what-is-a-reana-release) here, with the same name or rename as "Goal"?
- It would be helpful to move the table that shows where to release what (see [here](https://digital-repositories.web.cern.ch/digital-repositories/reana/releasing/#release))
- It would also be helpful to document the release order
    - Arguably, this is already documented now, as first we show Python package releases, then Docker and then Helm. To know which component is which (whether Python package, Docker or Helm) one can look up the previous table
- We would be missing the `reana-dev release-github`, if we want to keep the [release on GitHub](https://digital-repositories.web.cern.ch/digital-repositories/reana/releasing/#github).
    - So far we were releasing every component/package on GitHub, however, I don't think it is necessary as long as `reanahub/reana` is released on GitHub, which it is by using `reana-dev release-helm`
- Should [after release steps](https://digital-repositories.web.cern.ch/digital-repositories/reana/releasing/#after-release) be documented here? (follow get started again, forum and twitter posts....)

docs/development/releasing/index.md
```diff
@@ -1,42 +1,53 @@
 # Releasing
 
-## Helm Charts
+As prerequisites to release:
+
+- The repository that holds the code for the component/package to release has:
+  - A release commit as last commit in `master` branch
+  - The last commit in master is tagged.
```
mvidalgarcia, line 7: Indentation is not being applied properly here. I'm not sure if the double indentations is intended.

image

tiborsimko: We are not always releasing from `master`, e.g. `0.6.1` was released from `maint-0.6`.  So you can use a neutral language such as "from the release branch such as `maint-0.7` for bug fixes or `master` for new versions)"

tiborsimko: You can also add to the list as the final step "The code base was tested locally in production conditions" without specifyng too much what that means for now...

docs/development/releasing/index.md
```diff
@@ -1,42 +1,53 @@
 # Releasing
 
-## Helm Charts
+As prerequisites to release:
+
+- The repository that holds the code for the component/package to release has:
```
tiborsimko, line 5: ... for the component package that is to be released ...

docs/development/releasing/index.md
```diff
@@ -1,42 +1,53 @@
...
+
+**1.** Release the package(s)
 
 ```console
-$ reana-dev run-ci \
-    --admin-email john.doe@example.org \
-    --admin-password 123456 \
-    --component r-d-r-roofit \
-    --exclude-components=r-ui,r-m-broker
+$ reana-dev release-pypi -c <package-name>
```
tiborsimko, line 14: If somebody would copy paste, using syntax like "<foo>" will lead to shell redirections... Better to use other more directly followable example.

E.g. if the example emulates releasing new `r-server` package, you can use:

```console
$ cd ~/mysrc/reana-server
$ reana-dev docker-build -c . --no-cache
```

docs/development/releasing/index.md
```diff
@@ -1,42 +1,53 @@
...
-$ git commit
+**3.** Release the image(s)
+
+```console
+$ reana-dev release-docker -c <component-name>
 ```
 
-!!! info
-    The release will be created and added to the [REANA releases](https://github.com/reanahub/reana/releases) once the pull request is merged using the GitHub action [Helm Chart Releaser](https://github.com/marketplace/actions/helm-chart-releaser) in [`helm-releaser.yaml`](https://github.com/reanahub/reana/blob/master/.github/workflows/helm-releaser.yaml).
+Note: repeat `2.` and `3.` until you release all the images you need.
```
tiborsimko, line 37: I would remove this note as it is kind of obvious... Otherwise people can use be better learn about `-c r-component1 -c r-component1`...

docs/development/releasing/index.md
diegodelemos: > We are not always releasing from master, e.g. 0.6.1 was released from maint-0.6. So you can use a neutral language such as "from the release branch such as maint-0.7 for bug fixes or master for new versions)"

This statement, while true, doesn't work with current `reana-dev` that's why the non-neutral language. I wrongly assumed we would document future behaviours as we would implement them (in https://github.com/reanahub/reana/issues/413 in this case). Amending...

docs/development/releasing/index.md
tiborsimko: Hmm, I'm not sure it would be of interest to `docs.reana.io` web site visitors to see so much details... if it pops up in searching for usage guide for example... So if we do port it, then it had better be somehow separated from the user content, a bit "hidden" in another search index, or perhaps even presented in a different coloured style... So that people wont' be confused that when they search for helm (aiming to see how to install REANA on GCP), they would end up on a page explaining how to create Helm charts themselves.  The end user researchers installing REANA on their laptops could get really confused.

(Example: I'm not sure whether in the Kubernetes documentation one can find a corner that speaks about how to release Kubernetes itself? That would confuse quite a few users... Better keep it separated.)

I'm not sure how this Releasing section began here, but perhaps we could keep it separated? Either on DR docs, either on GitHub wiki... somewhere that it is not searched in common.  Or keep here but explore putting it under different section/search index? WDYT?

P.S. A minor comment about GitHub releases: this could indeed stay at the level of pushed tags only, especially for alpha/beta releases.  However, for production releases, we may want to populate the release news... Could be manually (since we don't do many of these per year yet), could be automated later via `release-github` indeed!

docs/development/releasing/index.md
tiborsimko: (Actually, the `--no-cache` step should have been done before, while testing all components together in production conditions, otherwise some docker layer caching problem could pop up at this (late) stage without it being fully tested...)

docs/development/releasing/index.md
tiborsimko: @diegodelemos What about removing the "Releasing" section from here and moving the content to `reana`'s GitHub Wiki, as we discussed IRL last week?

(Also, needs a rebase against latest master, but let's perhaps merge the other pending PR first.)

diegodelemos: Moved to GitHub wiki https://github.com/reanahub/reana/wiki/Releasing.

tiborsimko:
Please rebase and ready for merge :rocket:

PR #68, opened 2020-10-29
ci: added github actions workflow

audrium:
Here is an example of GH actions workflow: https://github.com/audrium/docs.reana.io/actions/runs/335964284


run-tests.sh
```diff
@@ -1,6 +1,42 @@
...
+}
+
+build_docs () {
+    mkdocs build -v
+    rm -rf site/
+}
+
+if [ $# -eq 0 ]; then
+    check_docstyle
+    build_docs
```
tiborsimko, line 31: Need also to include `check-shellscript` for "all-checks" target.

tiborsimko (approved):

PR #77, opened 2021-02-22
docs: added containers entry to advanced usage index page

audrium:


docs/advanced-usage/index.md
```diff
@@ -21,3 +21,7 @@
 - [Kerberos](access-control/kerberos) to access restricted resources
 - [VOMS-proxy](access-control/voms-proxy) to authenticate with a VOMS proxy certificate
 - [Private Docker registries](access-control/private-docker-registries) to access private images
+
+**Containers**
+
+- [docker](docker) - tips and tricks for using Docker environments
```
tiborsimko, line 27: (1) Please amend the text to "useful for using containerised Docker environments"

... in both instances, to go better with the rest that does not use `-`.

(2) Perhaps we should also use Docker label text with uppercase D, to go better with the rest.

tiborsimko: (1) Sorry, I left out the word "tips" in the phrase "useful tips for using Docker..."  :worried: Perhaps we can shorten it to say (1a) "tips for using Docker...", or (1b) "useful tips for Docker containerised environments", to avoid the double word "use" so shortly after each other.

(2) The link from the main index is not working, should be `containers/docker` including the directory.

tiborsimko (approved):

tiborsimko (approved):

PR #84, opened 2021-06-16
configuration: added configuring workflow scheduler page

audrium:
closes https://github.com/reanahub/reana-server/issues/357

### DIFF
--- a/docs/administration/configuration/configuring-scheduler/index.md
+++ b/docs/administration/configuration/configuring-scheduler/index.md
@@ -1,0 +1,17 @@
+# Configuring workflow scheduler
+
+As of version 0.8, REANA supports two different workflow scheduling strategies:
+
+- `fifo` - first-in first-out strategy starting workflows as they come.
+- `balanced` - a weighted strategy taking into account existing multi-user
+workloads and the [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph) complexity of incoming workflows.
+
+By default REANA is configured to use `fifo` workflow scheduler strategy.
+If you would like to change it, you can configure
+[`components.reana_server.environment.REANA_WORKFLOW_SCHEDULING_POLICY`](https://github.com/reanahub/reana/tree/master/helm/reana)
+Helm value and set the desired strategy accordingly.
+
+Note that it is also possible to change the workflow scheduler sleep time before re-inspecting
+the workflow queue and launching new workflows. By default, REANA uses a sleep time of 15 seconds.
+You can configure the Helm value [`components.reana_server.environment.REANA_SCHEDULER_REQUEUE_SLEEP`](https://github.com/reanahub/reana/tree/master/helm/reana).
+Please don't use values lower than 2 seconds; some delay is necessary for requeuing workflows in cluster overload situations.
--- a/docs/administration/configuration/index.md
+++ b/docs/administration/configuration/index.md
@@ -1,0 +1,4 @@
+# Configuration
+
+- [Configuring user access](configuration/configuring-access)
+- [Configuring workflow scheduler](configuration/configuring-scheduler)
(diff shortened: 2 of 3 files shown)


docs/administration/configuration/configuring-scheduler/index.md
```diff
@@ -0,0 +1,12 @@
...
+
+REANA supports two different workflow scheduling strategies:
+
+- `fifo` - first-in first-out strategy starting workflows as they come.
+- `balanced` - a weighted strategy taking into account existing multi-user
+workloads and the [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph) complexity of incoming workflows.
+
+By default REANA is configured to use `fifo` workflow scheduler strategy.
+If you would like to change it, you can configure
+[`components.reana_server.environment.REANA_WORKFLOW_SCHEDULING_POLICY`](https://github.com/reanahub/reana/tree/master/helm/reana)
```
tiborsimko, line 11: Hmm, shall we point to `master` or `maint-0.7`?  In the former case, we could advertise features that the installation might not have.  In the latter case, we would have to remember to update links before every new major version.

BTW I received a comment from one external installation that a feature advertised here was not available there.  Perhaps we could start putting "As of version 0.8, REANA supports..." qualifiers in the docs...

mvidalgarcia: There's no `maint-0.7` branch for this repo.

I think that what you mention it's a clear symptom that working in two separate branches and developing full features on both can cause these kinds of inconsistencies between what's in PROD and in `master`. IMO we should aim at a stronger CD strategy to have new features live as soon as possible.

tiborsimko: I meant to point to the `maint-0.7` branch in the `reana` repository, where the Helm variables live, not to include a new `maint-0.7` branch here in `docs.reana.io`. If we do this, and if we amend the head of `REANA.md` in each `reana` branch to clearly indicate version, then there should be no confusion as to which configuration option came when.

(Indicating a proper version number for every feature is fairly orthogonal to the question of deploying features as soon as possible. Basically I'm looking at a Sphinx-like `..versionadded` and `..versionchanged` technique to indicate clearly to the admins what is available since when... Hence the "(as of version 0.8)" comment...)

tiborsimko: @audrium Shall we add a few words about `REANA_SCHEDULER_REQUEUE_SLEEP`? Or perhaps not necessary?

audrium: +1 for "As of version 0.8", I think it's a bit cleaner than pointing to `maint-0.7` branch. And I hope 0.8 will soon be the default version ;)

I've added a few sentences about `REANA_SCHEDULER_REQUEUE_SLEEP`, @tiborsimko , please have a look

docs/administration/configuration/configuring-scheduler/index.md
```diff
@@ -0,0 +1,17 @@
...
+Helm value and set the desired strategy accordingly.
+
+Note that it is also possible to change time to wait between workflow consumption. By default REANA uses a 15 seconds delay.
```
tiborsimko, line 14: "time to wait between workflow consumption" may be hard to understand. What about something like:

Note that it is also possible to change the workflow scheduler sleep time before re-inspecting the workflow queue and launching new workflows. By default, REANA uses a sleep time of 15 seconds. You can configure the Helm value xxxxxxx. Please don't use values lower than 2 seconds; some delay in necessary for requeuing workflows in cluster overload situations.

audrium: This looks great :+1: . Updated accordingly

tiborsimko (approved):

Issue #91, opened 2021-09-04
docs: create a dedicated "For developers" page


VMois:
REANA is a complex system, making it easier for developers to start contributing to it will definitely help the project.
For this, we can create a dedicated "For developers" page to provide all necessary information for new contributors. The page should improve the developer experience. Example of the information that can appear on this page:

- how to set up a development environment and cluster ([here](http://docs.reana.io/administration/deployment/deploying-locally/#for-developers) is an existing resource but it is "hidden" under "Administration" page)
- guidelines for code formatting and how to contribute
- tips and tricks on how to debug the cluster or use `reana-dev` tool
- architecture of the REANA cluster and how different components interact

Examples of similar "For developers" pages:

- [Dask Distributed "Developer documentation"](https://distributed.dask.org/en/latest/communications.html)
- [React "Contribution" page](https://reactjs.org/docs/how-to-contribute.html)

We can start small by moving already existing resources for developers to a dedicated page. Thank you.

tiborsimko:
Two thoughts on this:

(1) Note that we used to have some developer-oriented pages on docs.reana.io but that we had discussed moving all developer-oriented information into GitHub Wiki. This was done for some pages, e.g. see "how to debug" https://github.com/reanahub/reana/wiki/Debugging

The motivation for separating researcher-level documentation from developer-level documentation is simple: not to make a deluge of information for researchers that might be searching main REANA docs for how to use (or install/administer) REANA. 
For example, if a search for some concept gives a lot of `reana-dev` hits, users would be confused since they are not seeing this (developer-only)wrapper. Having this on GitHub wiki allows to nicely separate concerns.

(See say https://github.com/reanahub/docs.reana.io/pull/73 and https://github.com/reanahub/docs.reana.io/issues/60 and others for examples of past activities in this direction.

(2) The most important tips and tricks for using `reana-dev` should be all documented in `reana-dev --help`, see the beginning of that command's output. Is this enough? Would you like to see there something else? We can definitely enrich it. Or have you been thinking of moving this kind of information fully to the developer wiki?  That might also be OK, but then we may need to think about keeping  the info in sync.  Since the number of contributors to REANA is rather small and well localised, as opposed to say React, keeping the reana-dev info in CLI help might do?

VMois:
(1) 
[...]

PR #95, opened 2021-09-23
docs: moves developers docs to GitHub Wiki

VMois:
- adds "For developers" page that links to GitHub Wiki
- removes "deploying locally => for developers" section

Relates to #91


docs/for-developers/.pages
```diff
@@ -0,0 +1,3 @@
+title: For developers
```
tiborsimko, line 1: I'm not sure we'd benefit from a new page that only points to the GitHub... What about keeping the single page, having "For researchers" as usual, and "For developers" at the end with a one-liner pointing to GitHub?

docs/for-developers/index.md
```diff
@@ -0,0 +1,4 @@
+# For developers
+
+If you are a developer and would like to contribute to the REANA cluster source code,
```
tiborsimko, line 3: If you are a developer and would like to install REANA locally and contribute to the REANA cluster code, please see [REANA wiki on GitHub](https://github.com/reanahub/reana/wiki/Development-installation).

docs/for-developers/.pages
VMois: I don't think "Deploying locally" is a proper page to direct developers that want to contribute to REANA. It is hidden in a not obvious place.

tiborsimko (approved):

PR #97, opened 2021-11-24
admin: add upgrading database section

mvidalgarcia:
addresses https://github.com/reanahub/reana/issues/591

### DIFF
--- a/docs/administration/deployment/deploying-at-scale/index.md
+++ b/docs/administration/deployment/deploying-at-scale/index.md
@@ -13,7 +13,7 @@ REANA can be easily deployed on large Kubernetes clusters using Helm. Useful for
     For example, [CERN uses CephFS](https://clouddocs.web.cern.ch/containers/tutorials/cephfs.html) as the default storage backend.
 
 !!! note
-    If you do not have any particular distributed file system in your Kubernetes cluster, you can easily [deploy an NFS network file system following our documentation](../../advanced-usage/storage-backends/nfs).
+    If you do not have any particular distributed file system in your Kubernetes cluster, you can easily [deploy an NFS network file system following our documentation](../../../advanced-usage/storage-backends/nfs/).
 
 ## Deploy
 
--- a/docs/administration/deployment/deploying-locally/index.md
+++ b/docs/administration/deployment/deploying-locally/index.md
@@ -17,9 +17,9 @@ $ firefox https://helm.sh/docs/intro/install/
 **2.** Deploy REANA cluster:
 
 ```console
-$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.7/etc/kind-localhost-30443.yaml
+$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.8/etc/kind-localhost-30443.yaml
 $ kind create cluster --config kind-localhost-30443.yaml
-$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.7/scripts/prefetch-images.sh
+$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.8/scripts/prefetch-images.sh
 $ sh prefetch-images.sh
 $ helm repo add reanahub https://reanahub.github.io/reana
 $ helm repo update
@@ -29,7 +29,7 @@ $ helm install reana reanahub/reana --namespace reana --create-namespace --wait
 **3.** Create REANA admin user:
 
 ```console
-$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.7/scripts/create-admin-user.sh
+$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.8/scripts/create-admin-user.sh
 $ sh create-admin-user.sh reana reana john.doe@example.org mysecretpassword
 ```
 
--- a/docs/administration/deployment/upgrading-db/index.md
+++ b/docs/administration/deployment/upgrading-db/index.md
@@ -1,0 +1,112 @@
+# Upgrading database
+
+If you are the REANA cluster administrator upgrading an existing REANA
+cluster deployment from one REANA release series to another, such as
+from 0.7.* to 0.8.0, there might be changes in the REANA database schema
+between the two versions. In these cases, it is necessary to run a
+database upgrade script using [`alembic`](https://alembic.sqlalchemy.org/en/latest/).
+
+The `reana-db` command-line tool is provided to ease the database
+upgrade tasks. The tool is included in the `REANA-Server` component.
+
+Two different procedures will be explained. The first, covers a quick
+database upgrade after a successful cluster upgrade. The second, covers
+a more detailed upgrade procedure, in case you want to have a better
+control or you need to troubleshoot some unexpected issues during the
+upgrade.
+
+## Quick upgrade procedure
+
+After you upgrade REANA by running `helm upgrade` successfully, run the
+following command:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -c rest-api -- reana-db alembic upgrade
+```
+
+The output will show the accomplished upgrades, for example:
+
+```console
+$ kubectl default exec -i -t deployment/reana-server -c rest-api -- reana-db alembic upgrade
+INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
+INFO  [alembic.runtime.migration] Will assume transactional DDL.
+INFO  [alembic.runtime.migration] Running upgrade  -> c912d4f1e1cc, Quota tables.
+INFO  [alembic.runtime.migration] Running upgrade c912d4f1e1cc -> ad93dae04483, Interactive sessions.
+INFO  [alembic.runtime.migration] Running upgrade ad93dae04483 -> 4801b98f6408, Job started and finished times.
+INFO  [alembic.runtime.migration] Running upgrade 4801b98f6408 -> f84e17bd6b18, Workflow complexity.
+INFO  [alembic.runtime.migration] Running upgrade f84e17bd6b18 -> 6568d7cb6710, storing full workflow workspace.
+```
+
+At this point, the upgrade of the database finished successfully and the
+procedure is done.
+
+## Detailed upgrade procedure
+
+First, you can check all the alembic `history` to list all the
+existing revisions:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -c rest-api -- reana-db alembic history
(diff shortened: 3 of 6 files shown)


docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,33 @@
...
+$ kubectl -n default exec -i -t deployment/reana-server -- reana-db alembic upgrade
+Defaulted container "rest-api" out of: rest-api, scheduler
+INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
+INFO  [alembic.runtime.migration] Will assume transactional DDL.
+INFO  [alembic.runtime.migration] Running upgrade  -> c912d4f1e1cc, Quota tables.
+INFO  [alembic.runtime.migration] Running upgrade c912d4f1e1cc -> ad93dae04483, Interactive sessions.
+INFO  [alembic.runtime.migration] Running upgrade ad93dae04483 -> 4801b98f6408, Job started and finished times.
+INFO  [alembic.runtime.migration] Running upgrade 4801b98f6408 -> f84e17bd6b18, Workflow complexity.
+INFO  [alembic.runtime.migration] Running upgrade f84e17bd6b18 -> 6568d7cb6710, storing full workflow workspace.
+```
```
tiborsimko, line 33: It would be good to add two new paragraphs with two more examples showcasing:

- `alembic history` so that people can see what has been applied previously on their systems;
- `alembic downgrade` so that people can go back in case something goes wrong with the deployment.

This may be useful for cluster administrators that are not familiar with Alembic.

P.S. And we could perhaps end up with referring people to the help:

```console
$ kubectl exec deployment/reana-server -c rest-api -- reana-db alembic --help 
```

mvidalgarcia: Added a user story following these steps:
1. upgrade
2. history
3. current
4. downgrade
5. current

docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,87 @@
...
+
+The `upgrade` command without any additional argument will attempt to
+upgrade the database to the latest revision. You can also pass a
+revision hash in the event that you just want to upgrade to a certain
+revision.
+
+Additionaly, you can check all the alembic `history` to list all the
```
tiborsimko, line 38: I think this "additionally" goes a bit counter the present story. The above story presented "usual upgrade" story, while this part starts "careful upgrade" story.

We have two options: 

- (a) introduce sections, since the page is getting long. The above could be entitled "**Quick upgrade procedure**" and the new one perhaps "**Detailed upgrade procedure** " and we start by explaining that if you want to have more control about the upgrade process, you start by checking which model version you have, the history, then do upgrade, then test and see, and if ever you'd like to downgrade this is how...

- (b)  We don't introduce sections and change the story to the "careful" one, as this will allow to explain everything smoothly in temporal progressive manner.  (E.g. CLI output won't "go back" as is the case now.)

WDYT? I guess former might be good, with first section being good for "quick impatient admins" and the section section may be good for those who'd like to understand more behind the scene.

Depending on which one we choos, the paragraph beginning "The `upgrade` command without any additional argument..." would move accordingly.

docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,87 @@
...
+INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
+INFO  [alembic.runtime.migration] Will assume transactional DDL.
+6568d7cb6710 (head)
+```
+
+In some exceptional cases you might need to `downgrade` to a previous
+revision. For example, due to a failure during the release upgrade
+process, or because you were testing an alpha version and you want to
+revert to the previous state. To do this, you need to specify the exact
+hash of the revision you want to downgrade to. This example shows a
```
tiborsimko, line 63: The following example shows...

docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,87 @@
...
+command to downgrade the database schema to a previous state, right
+after the `Job started and finished times` were introduced.
```
tiborsimko, line 65: Trailing `.` should be `:` as we are showing console output.

tiborsimko, line 65: right after the database change "Job started and finished times" was introduced:

docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,87 @@
...
+between the two versions. In these cases, it is necessary to run a
+database upgrade script using [`alembic`](https://alembic.sqlalchemy.org/en/latest/).
+
+The `reana-db` command-line tool is provided to ease the database
+upgrade tasks. The tool is included in the `REANA-Server` component.
+After you upgrade REANA by running `helm upgrade` succcessfully, you can
+run the following command:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- reana-db alembic upgrade
```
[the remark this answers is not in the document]
mvidalgarcia: Reminder added: https://github.com/reanahub/reana/pull/595#issuecomment-979978749

docs/administration/deployment/upgrading-db/index.md
mvidalgarcia: I guess (b) makes more sense, I'll try something and push it.

docs/administration/deployment/deploying-at-scale/index.md
```diff
@@ -13,7 +13,7 @@ REANA can be easily deployed on large Kubernetes clusters using Helm. Useful for
     For example, [CERN uses CephFS](https://clouddocs.web.cern.ch/containers/tutorials/cephfs.html) as the default storage backend.
 
 !!! note
-    If you do not have any particular distributed file system in your Kubernetes cluster, you can easily [deploy an NFS network file system following our documentation](../../advanced-usage/storage-backends/nfs).
+    If you do not have any particular distributed file system in your Kubernetes cluster, you can easily [deploy an NFS network file system following our documentation](../../../advanced-usage/storage-backends/nfs).
```
audrium, line 16: Link does not work for me. I get 404 error after pressing it. Appending "/" or "/index.md" solves the problem

docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,110 @@
...
+
+Two different procedures will be explained. The first, covers a quick
+database upgrade after a successful cluster upgrade. The second, covers
+a more detailed upgrade procedure, in case you want to have a better
+control or you need to troubleshoot some unexpected issues during the
+upgrade.
+
+## Quick upgrade procedure
+
+After you upgrade REANA by running `helm upgrade` succcessfully, run the
```
audrium, line 20: succcessfully -> successfully

docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,110 @@
...
+$ kubectl exec -i -t deployment/reana-server -c rest-api -- reana-db alembic upgrade f84e17bd6b18
+INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
+INFO  [alembic.runtime.migration] Will assume transactional DDL.
+INFO  [alembic.runtime.migration] Running upgrade 4801b98f6408 -> f84e17bd6b18, Workflow complexity.
+```
+
+In some exceptional cases you might need to `downgrade` to a previous
+revision. For example, due to a failure during the release upgrade
+process, or because you were testing an alpha version and you want to
+revert to the previous state. Imagine that something went wrong in the
```
audrium, line 85: Minor: this sentence which starts with "Imagine.." is kind of duplicate with a previous one IMO. Since in the previous sentence you already explained that one may want to downgrade in case of failure or testing alpha

docs/administration/deployment/upgrading-db/index.md
```diff
@@ -0,0 +1,110 @@
...
+```console
+$ kubectl default exec -i -t deployment/reana-server -c rest-api -- reana-db alembic upgrade
+INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
+INFO  [alembic.runtime.migration] Will assume transactional DDL.
+INFO  [alembic.runtime.migration] Running upgrade  -> c912d4f1e1cc, Quota tables.
+INFO  [alembic.runtime.migration] Running upgrade c912d4f1e1cc -> ad93dae04483, Interactive sessions.
+INFO  [alembic.runtime.migration] Running upgrade ad93dae04483 -> 4801b98f6408, Job started and finished times.
+INFO  [alembic.runtime.migration] Running upgrade 4801b98f6408 -> f84e17bd6b18, Workflow complexity.
+INFO  [alembic.runtime.migration] Running upgrade f84e17bd6b18 -> 6568d7cb6710, storing full workflow workspace.
+```
```
audrium, line 38: I guess here we could add one more sentence explaining users that the database at this point was migrated successfully, that it's done and that they don't need to do anything else, WDYT?

docs/administration/deployment/deploying-at-scale/index.md
mvidalgarcia: Strange.. it works for me, maybe it's a browser problem? I'll append `/` just in case.

audrium (approved):

docs/administration/deployment/deploying-locally/index.md
```diff
@@ -17,9 +17,9 @@ $ firefox https://helm.sh/docs/intro/install/
 **2.** Deploy REANA cluster:
 
 ```console
-$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.7/etc/kind-localhost-30443.yaml
+$ wget https://raw.githubusercontent.com/reanahub/reana/0.8.0/etc/kind-localhost-30443.yaml
 $ kind create cluster --config kind-localhost-30443.yaml
-$ wget https://raw.githubusercontent.com/reanahub/reana/maint-0.7/scripts/prefetch-images.sh
+$ wget https://raw.githubusercontent.com/reanahub/reana/0.8.0/scripts/prefetch-images.sh
```
tiborsimko, line 22: It is better to use `maint-0.8` rather than the fixed tag `0.8.0`, because we might have some bug fixes coming in the next days... In such a case, the instructions wouldn't work when we shall be in-between 0.8.0 and 0.8.1 point releases. Whilst having `maint-0.8` is safer, since it is moving target, and the committed bug fixes work "immediately" there, regardless of point releases. 

(That's basically why we went for `maint-x.y` there.)

mvidalgarcia: Updated and branch `maint-0.8` pushed to `reana`.

PR #100, opened 2022-01-06
kubernetes: add job_timeout docs section

VMois:
closes reanahub/reana-job-controller#343

### DIFF
--- a/docs/advanced-usage/compute-backends/kubernetes/index.md
+++ b/docs/advanced-usage/compute-backends/kubernetes/index.md
@@ -7,15 +7,14 @@ If `step` does not contain `compute_backend` specification, it will be executed
 on the default backend.
 
 ```yaml hl_lines="6"
-
-   # Serial example
-   ...
-   steps:
-      - name: reana_demo_helloworld_htcondorcern
-        environment: 'python:2.7-slim'
-        compute_backend: kubernetes
-        commands:
-            - python "${helloworld}"
+  # Serial example
+  ...
+  steps:
+    - name: reana_demo_helloworld_htcondorcern
+      environment: 'python:2.7-slim'
+      compute_backend: kubernetes
+      commands:
+        - python "${helloworld}"
 ```
 
 ## Custom memory limit
@@ -28,29 +27,25 @@ Read more about the expected memory values on [Kubernetes official documentation
 
 You can configure the `steps` in the respective specifications for Serial, Yadage, CWL and Snakemake workflow engines.
 
-### Serial workflow
-
-You can set `kubernetes_memory_limit` in every `step` of workflow specification.
+For **Serial**, you can set `kubernetes_memory_limit` in every `step` of workflow specification:
 
 ```yaml hl_lines="6"
-   ...
-   steps:
-      - name: reana_demo_helloworld_memory_limit
-        environment: 'python:2.7-slim'
-        compute_backend: kubernetes
-        kubernetes_memory_limit: '8Gi'
-        commands:
-            - python helloworld.py
+  ...
+  steps:
+    - name: reana_demo_helloworld_memory_limit
+      environment: 'python:2.7-slim'
+      compute_backend: kubernetes
+      kubernetes_memory_limit: '8Gi'
+      commands:
+        - python helloworld.py
 ```
 
-### Yadage workflow
-
-You can set `kubernetes_memory_limit` in every `step` under `environment.resources`.
+For **Yadage**, you can set `kubernetes_memory_limit` in every `step` under `environment.resources`:
 
 ```yaml hl_lines="19"
-   ...
-   stages:
-   - name: reana_demo_helloworld_memory_limit
+  ...
+  stages:
+    - name: reana_demo_helloworld_memory_limit
       dependencies: [init]
       scheduler:
         scheduler_type: 'singlestep-stage'
@@ -69,13 +64,11 @@ You can set `kubernetes_memory_limit` in every `step` under `environment.resourc
               - kubernetes_memory_limit: '8Gi'
 ```
 
-### CWL workflow
-
-You can set `kubernetes_memory_limit` in every `step` under `hints.reana`.
+For **CWL**, you can set `kubernetes_memory_limit` in every `step` under `hints.reana`:
 
 ```yaml hl_lines="7"
-   ...
-   steps:
+  ...
+  steps:
     first:
       hints:
         reana:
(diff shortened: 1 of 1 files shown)


VMois:
Looks like duplicated titles are not allowed in the same document. I will need to:

A) change a name from "Serial workflow" to something like "Specifying job timeout in Serial workflow"

B) merge "custom memory limit" and "job timeout" section

I would prefer option `A`, but I am open to suggestions. What do you think, @tiborsimko?

tiborsimko:
I agree that `A` is better. We can change subtitles to "Specifying custom memory limits for Yadage workfows" etc. Pros: this will make the new sections nicely anchorable/referenceable. Cons: at the expense of the subtitles being too long and the page structure having too many subtitles.

There is a `C` alternative where we can leave only top-level subtitles, i.e. custom memory limits and runtime job limits, and the various cwl/serial/snakemake/yadage examples will be simply presented with `**bold**`. Pros: this will give fewer sections. Cons: we won't have HTML anchors for linking to say the concrete CWL runtime example.  

Since our examples are short, I think the last con is not that huge, so `C` might be better than `B`. (Especially if we expand this section in the future with another configuration option added, or with another workflow engine added.) `C` might be more future-proof in this respect.

VMois:
I implemented point `C` so only top-level headings.

mvidalgarcia:
Needs rebase.

docs/advanced-usage/compute-backends/kubernetes/index.md
```diff
@@ -108,3 +100,87 @@ You can set `kubernetes_memory_limit` in every `rule` under `resources`.
         "docker://python:2.7-slim"
     ...
 ```
+
+## Job timeout
```
tiborsimko, line 104: `## Custom job timeouts` would play more nicely with "Custom memory limits"

(note also about plurals, may be more intuitive than singulars in titles, since people think usually about a workflow consisting of many jobs of diverse nature)

docs/advanced-usage/compute-backends/kubernetes/index.md
```diff
@@ -108,3 +100,87 @@ You can set `kubernetes_memory_limit` in every `rule` under `resources`.
...
+
+When the job exceeds the specified time limit, it will be terminated by Kubernetes and marked as failed.
```
tiborsimko, line 106: When a job

docs/advanced-usage/compute-backends/kubernetes/index.md
```diff
@@ -108,3 +100,87 @@ You can set `kubernetes_memory_limit` in every `rule` under `resources`.
...
+
+To set the job timeout, you can specify `kubernetes_job_timeout` in the specification of **each workflow step**.
```
tiborsimko, line 108: "you can declare ... in the specification"  to avoid the word repetition

docs/advanced-usage/compute-backends/kubernetes/index.md
```diff
@@ -108,3 +100,87 @@ You can set `kubernetes_memory_limit` in every `rule` under `resources`.
...
+Time is measured in **seconds**.
+
+Read more about the expected values and timeout logic on [Kubernetes official documentation](https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup)
```
tiborsimko, line 111: The K8s documentation prefer the term "job time limit" and some such. They don't really use "timeout". I guess we can use the former as it should be clear, let's simply reformulate a bit:

```
Read more about the job run time deadline limits in the _Kubernetes official documentation_.
```

Note also the trailing dot.

docs/advanced-usage/compute-backends/kubernetes/index.md
```diff
@@ -108,3 +100,87 @@ You can set `kubernetes_memory_limit` in every `rule` under `resources`.
...
+
+You can configure the `steps` in the respective specifications for Serial, Yadage, CWL and Snakemake workflow engines.
+
+For **Serial**, you can set `kubernetes_job_timeout` in every `step` of workflow specification.
```
tiborsimko, line 115: I would say let's end the paragraph with colon (`:`) which will better indicate that the code block below belongs to this phrase.  You can also change the `kubernetes_memory_limit` parts above to be the same...

docs/advanced-usage/compute-backends/kubernetes/index.md
```diff
@@ -108,3 +100,87 @@ You can set `kubernetes_memory_limit` in every `rule` under `resources`.
...
+    input:
+        helloworld=config["helloworld"],
+        inputfile=config["inputfile"],
+    params:
+        sleeptime=config["sleeptime"]
+    output:
+        "results/greetings.txt"
+    resources:
+        compute_backend="kubernetes",
+        kubernetes_job_timeout=60
```
tiborsimko, line 182: Hmm, I've been quoting the Snakefile value in my local tests `kubernetes_job_timeout="60"` :wink: but the integer is better.

docs/advanced-usage/compute-backends/kubernetes/index.md
```diff
@@ -108,3 +100,87 @@ You can set `kubernetes_memory_limit` in every `rule` under `resources`.
...
+Read more about the job run time deadline limits in the [Kubernetes official documentation](https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup).
+
+You can configure the `steps` in the respective specifications for Serial, Yadage, CWL and Snakemake workflow engines.
+
+For **Serial**, you can set `kubernetes_job_timeout` in every `step` of workflow specification:
+
+```yaml hl_lines="6"
+   ...
+   steps:
+      - name: reana_demo_helloworld_job_timeout
```
mvidalgarcia, line 120: I think there's too much indentation here, 2 spaces should do.

VMois: I will also fix the k8s memory limit. It has sometimes 3 spaces.

mvidalgarcia (approved):
Looking good!

mvidalgarcia:
Perhaps we should target this PR to `next` so the changes are not visible until we release 0.8.1?

VMois:
> Perhaps we should target this PR to `next` so the changes are not visible until we release 0.8.1?

Should I switch base to `next`? We have one sentence about releasing docs in [wiki](https://github.com/reanahub/reana/wiki/Releasing) (at the bottom).

mvidalgarcia:
> Should I switch base to `next`? We have one sentence about releasing docs in [wiki](https://github.com/reanahub/reana/wiki/Releasing) (at the bottom).

Yes, please. Otherwise changes will be visible automatically as we have CD configured for docs.

tiborsimko (approved):
Please rebase against latest `next`.

PR #106, opened 2022-01-17
docs: update reana-client CLI API

### LINKED ISSUE #104: docs: Reference CLI API
Check page https://docs.reana.io/reference/reana-client-cli-api/

E.g. `info` command is missing there.

We should update this page to latest version.

We might also consider leaving the body mostly empty and simply link to the RTFD web site for details, as we are doing for Python API. However, this has a disadvantage of users searching for some concept and not finding anything on `docs.reana.io`, so we should be careful here. Perhaps we can consider this later, after we populate tutorial parts (#26) which will touch many of these commands.  Still, an auto-generated reference guide might be of interest.

mvidalgarcia:
- Add missing `info` command
- Use double backticks everywhere so highlighted expressions are compatible with
both markdown (docs.reana.io) and reStructuredText (reana-client.readthedocs.io)

closes #104

### DIFF
--- a/docs/reference/reana-client-cli-api/index.md
+++ b/docs/reference/reana-client-cli-api/index.md
@@ -18,6 +18,7 @@ Quota commands:
   quota-show  Show user quota.
 
 Configuration commands:
+  info     List cluster general information.
   ping     Check connection to REANA server.
   version  Show version.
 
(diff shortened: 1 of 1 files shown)


audrium (approved):
Looks good to me!

mvidalgarcia:
> Looks good to me!

Hold on, there's still a related r-client PR missing 😁

Issue #99, opened 2021-12-20
List versions of CWL supported?


mr-c:

tiborsimko:
Hi @mr-c, we should definitely improve the REANA documentation in this respect.

Latest REANA 0.8 release series uses `cwltool==3.1.20210628163208`.

Eearlier REANA 0.7 release series used `cwltool==1.0.20191022103248`.

mr-c:
Thanks @tiborsimko . Using a particular version of `cwltool` as a library does not automatically mean that particular versions of CWL is fully supported, hence the requests to run the CWL conformance tests and report the results (even if only manually, once per release)

https://github.com/reanahub/reana-workflow-engine-cwl/issues/168

tiborsimko:
CC @audrium who has been dealting with running the CWL reference test suite lately. We'll upgrade the REANA documentation for each supported workfolw engine and we'll have a look specifically at reporting the CWL conformance test suite results afterwards.

audrium:
Hi @mr-c, we have just reran the conformance test suite on the latest REANA version and the results are as follows:
```
179 tests passed, 18 failures, 0 unsupported features
1 tool tests failed
```
The failures are documented in the [reana-workflow-engine-cwl](https://github.com/reanahub/reana-workflow-engine-cwl/issues) issues if you want to have a look.

We are planning to work on exposing these results in REANA documentation in the upcoming weeks

PR #107, opened 2022-01-17
docs: added CWL conformance results

audrium:
closes https://github.com/reanahub/docs.reana.io/issues/99

### DIFF
--- a/docs/running-workflows/supported-systems/cwl/index.md
+++ b/docs/running-workflows/supported-systems/cwl/index.md
@@ -1,3 +1,202 @@
 # CWL
 
-The Common Workflow Language standard originated in life science community.
+The Common Workflow Language standard for describing computational workflows originated in the life science community. Please see:
+
+- CWL project website: [https://www.commonwl.org](https://www.commonwl.org)
+- CWL user guide: [https://www.commonwl.org/user_guide](https://www.commonwl.org/user_guide)
+
+## Supported versions
+
+| REANA version          | ``cwltool`` version |
+| ---------------------- | ------------------- |
+| 0.8 release series     | 3.1.20210628163208  |
+| 0.7 release series     | 1.0.20191022103248  |
+| 0.6 release series     | 1.0.20190815141648  |
+| 0.5 release series     | 1.0.20181118133959  |
+| 0.3 release series     | 1.0.20180912090223  |
+| 0.2 release series     | 1.0.20180326152342  |
+
+## CWL v1.0 specification conformance results
+
+REANA 0.8.0 tested on 2022-01-12
+
+- 183 tests passed
+- 14 failures
+- 0 unsupported features
+
+<!-- markdownlint-disable no-inline-html  -->
+<details>
+    <summary>List of failed tests</summary>
+
+```plaintext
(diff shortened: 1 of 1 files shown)


docs/running-workflows/supported-systems/cwl/index.md
```diff
@@ -1,3 +1,178 @@
 # CWL
 
 The Common Workflow Language standard originated in life science community.
+
+Latest REANA 0.8 release series uses ``cwltool==3.1.20210628163208``.
```
tiborsimko, line 5: This could be better a table.

docs/running-workflows/supported-systems/cwl/index.md
```diff
@@ -1,3 +1,178 @@
...
+Eearlier REANA 0.7 release series used ``cwltool==1.0.20191022103248``.
+
+## CWL v1.0 specification conformance results
+
+REANA 0.8.0 tested on 2022-01-11
+
+- 183 tests passed
+- 14 failures
+
+```console
```
tiborsimko, line 16: Instead of console, let's make it a sublist of 14 failures.

docs/running-workflows/supported-systems/cwl/index.md
tiborsimko: We could expand the intro afterwards, independently of conformance tests results.

docs/running-workflows/supported-systems/cwl/index.md
tiborsimko: We could later think about adding some expand/collapse feature, so that readers would immediately see the overview only (183 passed, 14 fails), and they could click on "expand" to unroll the second sublist with all the failures.  It'll be probably nicer to do it with lists and sublists rather than with console areas.

audrium: Sublist takes a lot of space, that's why I went for `console` initially, since it renders it in a more compact way.
I've now added a expand/collapse feature:
 - had to move the list of failed tests outside the overview (183 passed, 14 fails) to the bottom due to the way it's being rendered in our docs
 - regular list was not rendered correctly in the expanded panel, so I had to use `plaintext` there
 - had to disable the linter warning to allow `html` tag usage

Let me know what you think about it

docs/running-workflows/supported-systems/cwl/index.md
```diff
@@ -1,3 +1,193 @@
 # CWL
 
 The Common Workflow Language standard originated in life science community.
```
tiborsimko, line 3: Please expand the introductory text a bit, for example:

```The Common Workflow Language standard for describing computational workflows originated in the life science community. Please see:

- CWL project website: https://www.commonwl.org/
- CWL user guide: https://www.commonwl.org/user_guide/

## Supported versions

The REANA platform implements CWL standard by means of using [cwltool](https://github.com/common-workflow-language/cwltool) reference implementation. The different REANA release series versions support different `cwltool` versions. The following table summarises the supported `cwltool` versions in each REANA release series:

... table ...
```

BTW please also dig out the past releases; it may be good to be complete here.

tiborsimko (approved):

PR #109, opened 2022-01-19
docs: Managing users and SSO

### LINKED ISSUE #102: docs: Managing users and SSO
The "Managing users" page https://docs.reana.io/administration/management/managing-users/ shows only local user scenario.  We should at least briefly mention several modes of operation: (i) local users created by admins; (ii) allowing users to create logins and asking for access; (iii) using SSO currently only with CERN.

This could be done by keeping the same page and introducing three "chapters" there.

We don't have to go into too much details, only to capture all these three modes, how to set them via Helm variables, so that we don't give false impression that SSO is not possible.

(This will pave the way for things like OIDC.)

(Test case: search the docs for SSO, it should give something reasonable instead of currently empty search results.)

mvidalgarcia:
closes #102


docs/administration/management/managing-users/index.md
```diff
@@ -34,22 +36,48 @@ Token for user aa37d63d-3186-45d5-aa40-5d221cb170bf (new.web.user@example.org) g
...
+![ui-sign-in](../../../images/ui-sign-in.png)
+
+As an administrator, you can configure REANA to disable the local users
+or to hide the sign up form. Please check the [Configuring user access](../../configuration/configuring-access) page to see how.
+
+## User registration via Single Sign-On
+
+Handling of users with Single Sign-On (SSO) is also possible for CERN
+deployments. This configuration can be combined with local users or used
+exclusively. When accesssing the UI you will see a page like this:
```
audrium, line 74: Minor: accesssing -> accessing

audrium (approved):
Looks good to me!

While reviewing I just discovered that links in main `configuration` page are not working. Same happens in production: https://docs.reana.io/administration/configuration/
While it's not directly related with PR, we could fix it here, since it's touching the `configuration` pages.. Fix can be done by removing `configuration/` from the links:
```
- [Configuring user access](configuring-access)
- [Configuring workflow scheduler](configuring-scheduler)
```

mvidalgarcia:
> While reviewing I just discovered that links in main `configuration` page are not working.

Good catch! Solved.

docs/administration/management/managing-users/index.md
```diff
@@ -34,22 +36,48 @@ Token for user aa37d63d-3186-45d5-aa40-5d221cb170bf (new.web.user@example.org) g
...
+
+## User registration via Single Sign-On
+
+Handling of users with Single Sign-On (SSO) is also possible for CERN
+deployments. This configuration can be combined with local users or used
+exclusively. When accessing the UI you will see a page like this:
+
+![ui-sso](../../../images/ui-sso.png)
+
+When clicking on "Sign in with SSO" you will be redirected to the
```
tiborsimko, line 78: Nice catch that we had a brief "Configuring user access section" already. Two thoughts upon seeing the changes:

- Looking at screenshots and the wording such as "when clicking on SSO..." etc, this is less for managing users than for explaining the whole user access in case SSO is enabled. Considering that, I think that the screenshots and the explanatory text may be better located in the "Configuring user access part" after all? We would present there the three options (default + disable sign-up + enable SSO) with screenshots and meaning what each is, plus how to enable/disable/configure by the admins.  And then here, on this page, we could basically just say a quick note for each command e.g. for `create` we could just say "If you have enabled SSO access, then user creation via CLI is not necessary since users will be created via SSO when they first log in".  This might read better in my eyes. WDYT?

- "Managing users" page stays a bit alone on the lhs menu amongst "Deployment" and "Configuration" menu sections.  Shall we add a new menu section "Management" and attach "Managing users" page under Management there, for consistency? Then we could add other sections such as "Managing disk space" there more easily in the future.

mvidalgarcia: - Makes sense, I've changed it. Let me know what you think.
- Agree, index Management page created.

tiborsimko (approved):

PR #121, opened 2022-04-07
docs: improve copy-to-clipboard

### LINKED ISSUE #120: ui: nicer copy-icon behaviour
When users copies CLI command instructions from [profile page](https://reana.cern.ch/profile), the leading `$ ` prompt signs are eliminated, so the examples are directly runnable.

This is not the case when copying CLI command examples from documentation pages [such as this one](https://docs.reana.io/getting-started/first-example/) where leading `$` prompt signs are inserted.

We should strive to amend the latter in order to behave more like the former in order to simplify user experience when copy-pasting runnable CLI execution examples from REANA documentation pages.

mdonadoni:
- Do not copy the terminal prompt `$` when copying console snippets
- Keep the copy-to-clipboard button only for some specific snippets

Closes #120

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -153,3 +153,14 @@ html .md-footer-meta .md-footer-meta__inner a {
 .md-typeset .admonition.note .admonition-title::before {
     color: #c8aeae;
 }
+
+/* show the copy-to-clipboard button only if .copy-to-clipboard is specified */
+.codehilite:not(.copy-to-clipboard) .md-clipboard {
+    display: none;
+}
+
+/* do not copy the terminal prompt `$` */
+pre > code > .gp {
+    -webkit-user-select: none;
+    user-select: none;
+}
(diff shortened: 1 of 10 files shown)


docs/stylesheets/extra.css
```diff
@@ -153,3 +153,14 @@ html .md-footer-meta .md-footer-meta__inner a {
...
 }
+
+/* show the copy-to-clipboard button only if .copy-to-clipboard is specified */
+.codehilite:not(.copy-to-clipboard) .md-clipboard {
+    display: none;
+}
+
+/* do not copy the terminal prompt `$` */
+pre > code > .gp {
+    -webkit-user-select: none;
```
mdonadoni, line 164: Note that on Safari the terminal prompts `$` are still being copied from multiline snippets, with the exception of the one in the first line (see this related [WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=80159))

VMois (commented):
Works nicely. Haven't looked into Safari. Left one small comment.

VMois (approved):

PR #127, opened 2022-05-31
docs: add page about user quotas

### LINKED ISSUE #125: quota: introduce exhaustive documentation
In view of introducing stricter disk quotas, we should create a nice documentation page for the end users.

This page should explain:

- how to check the quota usage and limits from CLI and Web
- what to do when quotas are reached: 
  - detecting workflows that consume the most space
  - deleting unnecessary past workflow runs
  - introducing shorter data retention period for big temporary ROOT files

Notes:

- The page should be for both CLI users (so containing console snippets) and Web UI users (so containing screenshots of functionalities that we develop, such as sort-by-most-used-disk-space-on-web-ui)
- We can start to prepare the page already now, and each now feature can simply enrich it when the feature is merged (both web ui, but also retention periods).
- There is a lot of text almost ready to be used by adapting past blog posts that were mentioning the quota feature, such as https://blog.reana.io/posts/2021/release-0.8.0/ which contain also user-oriented parts.

See also:

- https://github.com/reanahub/reana-ui/issues/271
- https://github.com/reanahub/reana-client/issues/607

mdonadoni:
Closes #125 

Note that the section about introducing shorter data retention periods to save disk space is not present in this PR, as this feature has not been implemented yet.


docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+The usage of the resources is also displayed on the profile page of the web UI,
+which you can access by clicking on the person-shaped icon in the top-right
+corner of the web interface:
+
+!["Your quota" section of the profile page](/images/ui-profile-your-quota.png)
+
+If the _"Your quota"_ section is not present, it means that user quotas are not
+enabled in your cluster!
```
VMois, line 46:
```suggestion
!!! note
    If the _"Your quota"_ section is not present, it means that user quotas are not enabled in your cluster!
```

Similar to https://docs.reana.io/administration/deployment/deploying-at-scale/. This will give you a nice "Note" box.

docs/advanced-usage/.pages
```diff
@@ -4,3 +4,5 @@ arrange:
     - storage-backends
     - code-repositories
     - access-control
+    - containers
```
VMois, line 7: *question:* why do you need to add `containers` here?

VMois (commented):
Nice job! Left some small comments. 

In addition, it looks like some features described here are still not deployed to `reana.cern.ch` so probably we will need to wait

docs/advanced-usage/.pages
mdonadoni: Because it is missing from the list of pages, we probably forgot to add it when the containers page was written

VMois: I was more curious why this change was there rather than the ordering of elements. Sorry for not being clear.

*suggestion:* It would be good to add it as a separate commit as this change is not part of user quotas docs.

But, if you already asked about ordering, usually we do alphabetic order in other places (code, etc.). I have checked some other docs pages and they are not in alphabetic order. But, I also don't think we should change the order here so just leave it as it is.

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,134 @@
...
+workflow"_:
+
+!["Delete workflow" menu option](/images/ui-delete-workflow.png)
+
+Make sure that _"Delete also workflow workspace"_ is checked before confirming
+the deletion of the workflow, otherwise the workspace will be kept intact and
+you will not free up any space:
+
+![Dialog to confirm the deletion of a workflow](/images/ui-delete-workflow-confirmation.png)
+
```
VMois, line 104: In the docs, you have usually provided both web UI and CLI options but, here, for deletion of workflows, it is missing. Maybe a short paragraph similar to this:

You can also delete a workflow with workspace using `reana-client`:

```
$ reana-client delete -w reana-demo-root6-roofit --include-workspace
```

For more details check `reana-client delete --help`

VMois (approved):
Nice work! The new page looks good. Especially I like the quality of the screenshots.

docs/advanced-usage/index.md
```diff
@@ -26,3 +26,7 @@
 **Containers**
 
 - [Docker](containers/docker) useful tips for Docker containerised environments
+
+**User Quotas**
+
+- [User Quotas](user-quotas) to deal with limits on the resources available
```
tiborsimko, line 32: - "User quotas" (BrE rather than AmE spelling of title word case)
- to inspect CPU and Disk resource quota limits and manage their use

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
+# User quotas
+
+Since
+[REANA 0.8.0](https://blog.reana.io/posts/2021/release-0.8.0/#cpu-and-disk-quota-accounting),
```
tiborsimko, line 4: Since
[REANA 0.8.0](https://blog.reana.io/posts/2021/release-0.8.0/#cpu-and-disk-quota-accounting), the cluster administrators have a possibility to set user quotas to limit the amount of disk space and CPU time available to execute workflows. This page explains how to inspect your limits and manage their use.

(taking user-oriented perspective, rather than admin-oriented perspective, in the narrative)

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
...
+it is possible to limit the amount of disk space and CPU time available to
+execute workflows with the help of user quotas.
+
+## Check usage and quota limits
+
+User quotas are optional features of REANA which might not be enabled in your
```
tiborsimko, line 10: ... which may or may not be enabled in your cluster.  You can check which resources are enabled in your cluster:

```console
$ reana-client quota-show --resources
cpu
disk
```

You can check the limits imposed by the cluster administrator as follows:

```console
$ reana-client quota-show --resource disk --report limit -h
No limit.
$ reana-client quota-show --resource cpu --report limit -h
1h
```

We see that this particular REANA example instance does not impose any limits on the disk usage; however the total runtime for all your workflows is limited to 1 hour only.

(Comment: this example is a bit unusual, because there would usually be some disk limit, and moreover the examples below are showing how to delete big files, which implies there is some disk usage quota.  (So the page narrative is a bit broken.) Also, the quota of 1h is too small for physics usage, so people may be confused whether this is for one run or for all runs. We could perhaps bump the CPU quota numbers, but hopefully people will understand that this is just an example, since I agree that it would be good to show in the docs also "unlimited" option! Hmm, perhaps we could show the disk quota limit to keep the page narrative, and show in a new section that cluster admins may also select "unlimited" so that you wolud see "No limit" etc. WDYT? We can muse IRL.)

You can inspect your current usage of the resources as follows:

```console
$ reana-client quota-show --resource disk -h
192 KiB
$ reana-client quota-show --resource cpu -h
1m 19s out of 1h used (2%)
```

(Comment: it looks a bit weird to show values without units, so I think we can focus on `-h` directly, or else always show both examples. The current example "192512 used" is very ambiguous)

(and now we can transition to the Web UI section...)

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
...
+1m 19s out of 1h used (2%)
+```
+
+The usage of the resources is also displayed on the profile page of the web UI,
+which you can access by clicking on the person-shaped icon in the top-right
+corner of the web interface:
+
+!["Your quota" section of the profile page](/images/ui-profile-your-quota.png)
+
+!!! note
```
tiborsimko, line 55: If the _"Your quota"_ section is not present on your profile page, it means that the cluster administrators did not enable user quotas feature on this REANA instance.

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
...
+    If the _"Your quota"_ section is not present, it means that user quotas
+    are not enabled in your cluster!
+
+## What to do when quotas are exceeded
+
+If you exceed your user quota of one of the resources, you will not be able to
+execute any new workflow until the resource usage will be back under the limit.
+In particular, if you have exceeded your CPU quota, there is nothing you can do
+except asking the administrators of your REANA instance to increase your CPU
+quota limit. However, if you have consumed all the available disk space, you can
```
tiborsimko, line 65: However, if you have consumed all the available disk space, you can download your data and either completely delete some past unnecessary workflows, or else remove the largest files, to get your disk space usage under quota limits.

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
...
+free up some of it by either deleting some past workflows or by removing some of
+their files.
+
+### Find workflows consuming the most disk space
+
+The easiest way to find out which workflows are using the most disk space is
```
tiborsimko, line 71: (Not sure about "easiest", it could depend on the user! Also, in the section above on this page, the CLI was presented before Web UI, so we may want to use same order here perhaps for consistency. (Or we could invert the order also above to have Web UI before CLI UI.) I assume we'd keep the former, hence suggested changes below.)

Using command line, you can use `reana-client list` command to also include the disk usage numbers for each workflow run by means of `--include-workspace-size` option, and have the list sorted by the most disk-space-consuming workflows first:

```console
$ reana-client list --include-workspace-size --sort size
NAME                      RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS     SIZE
reana-demo-root6-roofit   2            2022-05-30T12:50:46   2022-05-30T12:50:58   2022-05-30T12:51:13   finished   180224
reana-demo-helloworld     1            2022-05-30T11:59:48   2022-05-30T12:00:00   2022-05-30T12:00:15   finished   12288
reana-demo-root6-roofit   1            2022-05-30T11:59:58   2022-05-30T12:00:09
```

Using web interface, you can go to the REANA home page and select...

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+!["Delete workflow" menu option](/images/ui-delete-workflow.png)
+
+Make sure that _"Delete also workflow workspace"_ is checked before confirming
+the deletion of the workflow, otherwise the workspace will be kept intact and
+you will not free up any space:
+
+![Dialog to confirm the deletion of a workflow](/images/ui-delete-workflow-confirmation.png)
+
+You can also delete a workflow using `reana-client`, specifying the
```
tiborsimko, line 105: (Ditto here, we would put "Using command line, ..." CLI part before "Using web interface, ..." Web UI part.)

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
...
+Once you have identified the heaviest workflows, you can also find which are the
+biggest files in their workspaces.
+
+From the web UI, open the details page of the workflow, move to the
+_"Workspace"_ tab and click on _"Size"_ to sort the files in ascending or
+descending order of size:
+
+![Sort the files in the workspace by decreasing "Size"](/images/ui-workspace-sort-by-size.png)
+
+You can also achieve the same result by using `reana-client` together with
```
tiborsimko, line 126: (Ditto here, we would put "Using command line, ..." CLI part before "Using web interface, ..." Web UI part.)

docs/advanced-usage/user-quotas/index.md
```diff
@@ -0,0 +1,144 @@
...
+`--include-workspace` option to delete all the workflow's files:
+
+```console
+$ reana-client delete -w reana-demo-root6-roofit.2 --include-workspace
+==> SUCCESS: reana-demo-root6-roofit.2 has been deleted
+```
+
+See `reana-client delete --help` for more details.
+
+### Find and delete big files
```
tiborsimko, line 115: Perhaps call them "large files", "largest files" here and elsewhere in the text body. Might read nicer?

docs/advanced-usage/user-quotas/index.md
mdonadoni: > (Comment: this example is a bit unusual, because there would usually be some disk limit, and moreover the examples below are showing how to delete big files, which implies there is some disk usage quota. (So the page narrative is a bit broken.)

I have added a limit of 500 GiB on the disk usaget and a "Note" box to show the output of reana-client when there are no limits.

> Also, the quota of 1h is too small for physics usage, so people may be confused whether this is for one run or for all runs.

The CPU time limit is now 10000 hours.

docs/advanced-usage/user-quotas/index.md
mdonadoni: Each section now shows the instructions for the CLI before the ones for the web interface, with some little changes to the old text and to the proposed changes

tiborsimko (approved):

PR #131, opened 2022-06-29
docs: add page about TLS certificates

### LINKED ISSUE #130: docs: administering SSL certificates
REANA 0.9 changes the way SSL certificates are generated in order to fix opening notebook sessions with latest Traefik.  (https://github.com/reanahub/reana-workflow-controller/issues/457)

We are now creating default Traefik TLS certificate valid for 90 days. We should enrich admin-oriented documentation to say how they can generate their own certificates valid longer and upload them via secrets, or how they can use Let's Encrypt for TLS certificate generation and renewal.

Stems from https://github.com/reanahub/reana/pull/650#pullrequestreview-1020259226

mdonadoni:
Closes #130


docs/administration/configuration/configuring-tls-certificates/index.md
```diff
@@ -0,0 +1,66 @@
...
+`reana-custom-tls-secret`:
+
+```{ .console .copy-to-clipboard }
+$ kubectl create secret tls reana-custom-tls-secret --cert=./cert.crt --key=./cert.key
+```
+
+Finally, you must set the necessary Helm values:
+
+- `ingress.tls.self_signed_cert` should be set to `false`, so that the default
+  self-signed certificate is not generated
```
tiborsimko, line 55: ... is not generated; (to end with semi-colon)

docs/administration/configuration/configuring-tls-certificates/index.md
```diff
@@ -0,0 +1,66 @@
...
+- `ingress.tls.secret_name` should be set to the name of the Kubernetes secret
+  containing the certificate, in this case `reana-custom-tls-secret`
```
tiborsimko, line 57: ...`reana-mycert-secret`. to end with a dot.

docs/administration/configuration/configuring-tls-certificates/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Automatic certificate issuance
+
+Depending on how your cluster is configured, you might be able to let Kubernetes
+handle the issuance and configuration of TLS certificates. Different ingress
+controllers require different configurations, but sometimes you might need to
```
tiborsimko, line 63: This section reads a bit cryptic. I guess we would have to provide some Let's Encrypt tentative example to make it clearer to the reader.  Something like:

You may also be able to let Kubernetes handle the issuance and configuration of TLS certificates so that the process would be fully automatic. Different ingress controllers may require different configurations, but the setup usually requires to annotate the ingress object. For example, in order to use the Let's Encrypt certificate service with Traefik ingress controller, it would benecessary to discover and edit the ingress config map:

```
$ # discover ingress 
$ kubectl get configmaps --all-namespaces | grep ingress
$ # edit config map to input the acme section as listed below
$ kubectl edit configmaps -n kube-system ingress-traefik
$ # delete ingress pod to trigger certificate issuance 
$ kubectl -n kube-system delete $(kubectl -n kube-system get pod -o name | grep traefik)
```

where the newly added acme section reads as follows:

```
    [acme]
      email = "john.doe@example.org"
      storage = "acme.json"
      entryPoint = "https"
      ACMELogging = true
    [acme.tlsChallenge]
    [[acme.domains]]
      main = "reana.example.org"
```

tiborsimko (approved):

Issue #147, opened 2023-01-18
docs: add section for using a different ingress controller (traefik/nginx)


goseind:
When working with `traefik` as an ingress controller in one of our clusters (ref https://github.com/EOSC-Future/reana-cnaf-demo), we experienced some issues and thus decided to go with `nginx`, with this issue we'll add a section about ingress configuration including an example code snipped for `nginx`.

Further thoughts would be to allow the specification of `nginx` instead of `traefik` in the helm chart.

PR #148, opened 2023-01-18
docs: add configuring ingress controller nginx example

goseind:
Please see the description for this PR in issue #147


goseind:
@tiborsimko let me know if you're ok with the content so we can approve and merge it onto the 📝

PR #149, opened 2023-01-24
docs: add configuration of storage volumes

mdonadoni:


docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+      volume_size: 200
+      access_modes: ReadWriteMany
+    ```
+
+### HostPath
+
+!!! warning
+    Using HostPath volumes is not recommended, as they present many security risks. See the related [Kubernetes documentation page](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) for more information.
+
+REANA also supports using directories already present on the host node's filesystem. Note that this approach is not suitable for multi-node deployments, but only for local single-node instances of REANA.
```
tiborsimko, line 73: ... but only for local single-user single-node instances of REANA, for example during personal development.

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+
+To use a local directory as a storage volume, set the [`*_storage.backend`](https://github.com/reanahub/reana/tree/master/helm/reana) Helm value to `hostpath` and [`*_storage.hostpath.root_path`](https://github.com/reanahub/reana/tree/master/helm/reana) to the path of the directory you would like to use:
```
tiborsimko, line 75: Again the comment `*_storage` could be simply said `shared_storage` for better readability.

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+
+```{ .yaml .copy-to-clipboard }
+shared_storage: # or `infrastructure_storage`
+  backend: hostpath
+  hostpath:
+    root_path: "/var/my-reana-directory"
```
tiborsimko, line 81: We use "myanalysis", "mydata" etc usually, so let's say "/opt/myreana"  perhaps?

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+    ```{ .yaml .copy-to-clipboard }
+    shared_storaged: # or `infrastructure_storage`
+      backend: nfs
+      volume_size: 200
+      access_modes: ReadWriteMany
+    ```
+
+### HostPath
+
+!!! warning
```
tiborsimko, line 70: Shall we put the warning _after_ the first paragraph explaining what host path is and what it is suitable for?

(IOW, invert order of 1st-2nd paragraph?)

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+
+```{ .yaml .copy-to-clipboard }
+shared_storage: # or `infrastructure_storage`
+  backend: hostpath
+  hostpath:
+    root_path: "/var/my-reana-directory"
+```
+
+## Migration to new infrastructure volume
+
```
tiborsimko, line 85: It'll be good to preface the "Danger" part by explaining what this is about.  We can add a paragraph like:

REANA deployment chart uses the single `shared_storage` volume by default. If you have an existing cluster with many users with heavy disk space needs, and you would like to introduce the additional `infrastructure_storage` volume to better isolate the infrastructure component storage needs from user workloads, you can proceed as follows.

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+!!! note
+    This guide requires to upgrade REANA a few times by using the `helm` tool. To make sure that you are running the correct commands, you can use the [`helm-diff`](https://github.com/databus23/helm-diff) plugin to inspect the changes that will be made to REANA's deployment.
+
+1. Make sure REANA is already updated to version 0.9.0 or greater. In the following steps, we will refer to this version as `$VERSION`.
+2. Disable the scheduling of new workflows by setting `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` to zero.
+
+    ```{ .console .copy-to-clipboard}
+    $ helm upgrade reana reanahub/reana --version $VERSION \
+        --values my-values.yaml \
```
tiborsimko, line 97: `myvalues.yaml` here and elsewhere

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+    This procedure is delicate and it can lead to data loss. Carefully read the following instructions and, if possible, try them out on a test instance of REANA to make sure that everything works.
+
+!!! note
+    This guide requires to upgrade REANA a few times by using the `helm` tool. To make sure that you are running the correct commands, you can use the [`helm-diff`](https://github.com/databus23/helm-diff) plugin to inspect the changes that will be made to REANA's deployment.
+
+1. Make sure REANA is already updated to version 0.9.0 or greater. In the following steps, we will refer to this version as `$VERSION`.
+2. Disable the scheduling of new workflows by setting `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` to zero.
+
+    ```{ .console .copy-to-clipboard}
+    $ helm upgrade reana reanahub/reana --version $VERSION \
```
tiborsimko, line 96: Perhaps put the `--version` part onto a new line, so that we have one-line-one-value philosophy?

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+        --values my-values.yaml \
+        --set components.reana_server.environment.REANA_MAX_CONCURRENT_BATCH_WORKFLOWS=0
+    ```
+
+3. Wait for all the running workflows to finish. You can run `kubectl get pods` to make sure that workflows are not running.
+4. Update `my-values.yaml` to configure the new infrastructure volume, as explained in the [Configuration](#configuration) section.
+
```
tiborsimko, line 103: Perhaps put a full example for CephFS, so that people may something to copy/paste from.

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+## Configuration
+
+Volumes can be configured with two Helm values:
+
+- [`shared_storage`](https://github.com/reanahub/reana/tree/master/helm/reana) holds the configuration for the shared filesystem needed by REANA.
+  It is mandatory to configure this volume.
+- [`infrastructure_storage`](https://github.com/reanahub/reana/tree/master/helm/reana) configures the second optional volume used by the database and the message broker.
+  If not configured, the shared storage volume is used instead.
+  It has the same settings as [`shared_storage`](https://github.com/reanahub/reana/tree/master/helm/reana).
```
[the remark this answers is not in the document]
tiborsimko: (Second thought: since we would like to promote best practices here, perhaps we could always copy the `shared_storage` example and the `infrastructure_storage` example inside the same snippet?

This will allow us to:

- advertise best practices for "scaling out";
- advertise the sizes that one is huge one is smaller;
- and, the section about migrating from one volume to two volumes will make more sense.  (As in, the above was recommended way, but if you have an older cluster, you can upgrade as follows... kind of thinking.)

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+To use a local directory as a storage volume, set the [`*_storage.backend`](https://github.com/reanahub/reana/tree/master/helm/reana) Helm value to `hostpath` and [`*_storage.hostpath.root_path`](https://github.com/reanahub/reana/tree/master/helm/reana) to the path of the directory you would like to use:
+
+```{ .yaml .copy-to-clipboard }
+shared_storage: # or `infrastructure_storage`
+  backend: hostpath
+  hostpath:
+    root_path: "/var/my-reana-directory"
+```
+
+## Migration to new infrastructure volume
```
tiborsimko, line 84: Hmm, the title may not be self-understandable. (Migration from where?)

Perhaps this part could be better separated out into a blog post?  Or we amend the title to something longer like "Separating infrastructure storage volume".

mdonadoni: I personally like having this migration guide in the docs instead of in a blog post, so I have changed the title to "Separating infrastructure storage volume"

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
...
+    provisioner: manila-provisioner
+    type: "Geneva CephFS Testing"
+    availability_zone: nova
+    os_secret_name: os-trustee
+    os_secret_namespace: kube-system
+    cephfs_os_share_id: <cephfs-share-id>
+    cephfs_os_share_access_id: <cephfs-share-access-id>
+```
+
+For more information regarding CERN deployments, see the related [Kubernetes@CERN documentation page](https://kubernetes.docs.cern.ch/docs/storage/fileshares/#using-existing-shares).
```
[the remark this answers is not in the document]
mdonadoni: I have removed it

docs/administration/configuration/configuring-storage-volumes/index.md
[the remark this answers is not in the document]
mdonadoni: I don't like too much repeating two times the same example for both the shared and the infrastructure volume, not only because this makes the page longer, but also because I think it gives the false impression that both the shared and the infrastructure storage need to have the same backend (e.g. both of them as CephFS shares).

I have reformulated the last phrase a bit:
"In the configuration examples below we'll use the `shared_storage` volume, but the `infrastructure_storage` volume can be configured in exactly the same way. Note that typically, the shared storage volume can be huge (even tens of terabytes), while the shared storage volume can be as small as a few gigabytes, depending also on whether you are using an external database or not."

docs/administration/configuration/configuring-storage-volumes/index.md
```diff
@@ -0,0 +1,194 @@
+# Configuring storage volumes
+
+As described in the [Deploying at scale](../../deployment/deploying-at-scale/index.md) documentation page, REANA needs a shared filesystem to store the workflows' workspaces.
+It is also possible to configure a second storage volume to be used by REANA's infrastructure components, such as the database and the message broker.
+
+## Configuration
+
+Volumes can be configured with two Helm values:
+
+- [`shared_storage`](https://github.com/reanahub/reana/tree/master/helm/reana) holds the configuration for the shared filesystem needed by REANA.
```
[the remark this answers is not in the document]
mdonadoni: Should I also remove the links in the rest of the page? e.g. `shared_storage.backend`, `shared_storage.cephfs`, `shared_storage.hostpath.root_path`, etc.

docs/advanced-usage/index.md
```diff
@@ -9,7 +9,6 @@
 **Storage backends**
 
 - [EOS](storage-backends/eos) for reading from and publishing on EOS object storage system
-- [NFS](storage-backends/nfs) for deploying NFS network file system
 - [Shared storage](storage-backends/shared-storage) for sharing results between workflow steps
```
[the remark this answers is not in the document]
mdonadoni: I have updated the "Shared storage" page. I have not changed the page hierarchy for now

docs/administration/configuration/configuring-storage-volumes/index.md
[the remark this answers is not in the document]
tiborsimko: Yes, if they point to the same target, then I think it's better to remove them. It's a good usability principle to have different links pointing to different external targets. Might come from old NN usability guidelines, similar to https://www.nngroup.com/articles/duplicate-links/

docs/administration/configuration/configuring-storage-volumes/index.md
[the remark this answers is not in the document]
tiborsimko: Yes I agree that my suggestion would seem to say that using different backends for different shared volumes is not possible, which would be very misleading... OTOH, if we don't display a full "scalable setup", then people might simply not use it.  Perhaps we could solve it by having a new subsection with a "complete example" which could actually show the infra volume being on NFS and the runtime shared volume being on CephFS as one possible robust setup? (Could be done also later.)

docs/administration/configuration/configuring-storage-volumes/index.md
tiborsimko: That works too, now that we have an introductory paragraph saying when/why this would be needed.

tiborsimko (approved):

PR #153, opened 2023-03-08
cli-api: fix typo in reana-client du command help

giuseppe-steduto:


tiborsimko (commented):
A cosmetic comment to the commit headline. It's not a fix for `du` command _specification_, but help message.  So I'd rather say something like:

```
cli-api: fix typo in reana-client du command help
```

P.S. We can start using `page: ...` rather than somewhat generic `docs: ...` prefix in the commit log headline, since in this repository almost everything would be `docs: ...` anyway.  We used that for some commits, but were not fully consistent yet...

PR #166, opened 2023-06-05
Create inline links for dependency install doc websiter

meliache:
I see not much value in providing a code block how to open the documentation websiter from the CLI and I would not assume everyone uses firefox. This seems the right place for good old hyperlinks to me.


tiborsimko (approved):
Thanks! 

BTW I fixed the markup formatting so that it would play nicely with the `mkdocs` driven web display, and I have removed the unnecessary merge commit in the feature branch.

PR #167, opened 2023-06-08
docs: add configuration of external access

mdonadoni:


docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+REANA creates a single Ingress object by default, which can be customized with the following [Helm values](https://github.com/reanahub/reana/tree/master/helm/reana):
+
+- `ingress.enabled` enables or disables the creation of Ingresses
+- `ingress.annotations` is used to define which custom annotations should be applied to the Ingress resource
+- `ingress.tls` can be used to configure a TLS certificate (see [Configuring TLS certificates](../configuring-tls-certificates/))
+
+You can also specify additional Ingresses with the `ingress.extra` Helm value.
+Each additional Ingress can be configured in the same way as the default one, with the addition of `ingress.extra[].name`.
+
+As an example, this is how you would configure Traefik by creating two Ingresses, one to handle HTTPS traffic and the other to redirect HTTP requests to HTTPS.
```
tiborsimko, line 17: ... to HTTPS:

(indicating that the topic is not finished and that the code snippet below will exemplify it)

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Manual setup
+
+If you would like to manually manage Ingresses, you should set `ingress.enabled` to false to disable the automatic creation of Ingresses.
+In the same way, you can disable the internal Traefik instance by setting `traefik.enabled` to false.
+
+Your custom Ingress must redirect all the requests to `/api` and `/oauth` to the `<helm-release-prefix>-server` Service, while all the rest should be sent to `<helm-release-prefix>-ui`
+You can find an example of custom Ingress below.
+
+??? note "Example of custom Ingress"
```
tiborsimko, line 48: Do we want to hide the box?  I think we can expand it by default...

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+                  port:
+                    number: 80
+          host: <hostname>
+    ```
+
+## Reverse proxy
+
+You might be interested in placing a (reverse) proxy in front of REANA, but some additional configuration might be needed to make REANA work correctly.
+
+Let's consider a setup in which a generic proxy handles connections to REANA, like so:
```
tiborsimko, line 92: Let us consider...  (slightly more neutral that contracted "let's")


... handles connections to REANA:

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
+# Configuring external access
+
+REANA manages external access with the use of Kubernetes [Ingresses](https://kubernetes.io/docs/concepts/services-networking/ingress/).
```
tiborsimko, line 3: ... manages external traffic to the cluster with the use of...

Perhaps?  (The access nomenclature also depends on how we solve the page naming issue mentioned above.)

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+    traefik.ingress.kubernetes.io/router.tls: "true"
+  extra:
+    - name: http
+      annotations:
+        kubernetes.io/ingress.class: traefik
+        traefik.ingress.kubernetes.io/router.entrypoints: http
+        traefik.ingress.kubernetes.io/router.middlewares: kube-system-redirect-scheme@kubernetescrd
+```
+
+## Manual setup
```
tiborsimko, line 40: The title could be "Setting up manual Ingresses" to play better consistency with the previous title name.

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
+# Configuring external access
+
+REANA manages external access with the use of Kubernetes [Ingresses](https://kubernetes.io/docs/concepts/services-networking/ingress/).
+The Helm chart automatically deploys Traefik as an [Ingress Controller](https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/) and creates the necessary Ingress resources, but this can be changed if you wish to deploy your own Ingress Controller or your own Ingresses.
+
+## Configuring generated Ingresses
```
tiborsimko, line 6: The title could be "Configuring default Ingresses" perhaps?

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+              pathType: Prefix
+              backend:
+                service:
+                  name: <helm-release-prefix>-ui
+                  port:
+                    number: 80
+          host: <hostname>
+    ```
+
+## Reverse proxy
```
tiborsimko, line 88: The title could be "Using reverse proxy" for better analogy with the above titles?

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+You might be interested in placing a (reverse) proxy in front of REANA, but some additional configuration might be needed to make REANA work correctly.
```
tiborsimko, line 90: Again avoiding but: You may be interested in placing a (reverse) proxy service in front of REANA. This requires additional configuration to make REANA work correctly.

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+Let's consider a setup in which a generic proxy handles connections to REANA, like so:
+
+```text
+          +---------+    +---------+
+Client -> |  Proxy  | -> |  REANA  |
+          +---------+    +---------+
+```
+
+In this case, REANA has no way to know how the client connected to the server, as REANA will only see connections coming from the proxy.
```
tiborsimko, line 100: ... has no way to know how the client connected to the server and which protocol the client used, since...

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+If the proxy also performs TLS termination, all the connections made to REANA will be over HTTP, even if the client connected over HTTPS.
```
tiborsimko, line 101: ... all the connectionsto REANA will be carried out over HTTP protocol, even if ...

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+This makes it impossible to construct correct URLs on the server-side, as the protocol to use is not known.
```
tiborsimko, line 102: This makes it impossible to construct correct absolute URLs on the server-side, since the traffic protocol is not known.

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+Even though not standard, the `X-Forwarded-*` family of HTTP headers can be used to preserve the details of the original connection and make them known to REANA's web server.
```
tiborsimko, line 104: In these situations, you can make use of a non-standard `X-Forwarded-*` family of HTTP headers to preserve the details about the original connection.

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+To fix the particular issue presented before, it is important that proxies correctly set the `X-Forwarded-Proto` HTTP header.
```
tiborsimko, line 105: It is  important to configure your proxy to provide correctly the `X-Forwarded-Proto` header.

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+Note that the provided Traefik Ingress Controller acts as a proxy, but it correctly sets these headers by default.
+
+!!! warning
+
+    Using `X-Forwarded-*` headers can create security issues.
+    You must make sure that the `X-Forwarded-*` headers provided by the client are either discarded or overwritten by a proxy you trust.
+    See for example the [security concerns for X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#security_and_privacy_concerns).
+
+Let's consider another setup in which HAProxy sits in front of the provided Traefik instace, like this:
```
tiborsimko, line 114: Let us now consider another setup in which HAProxy sits in front of a separate Traefk instance:

(note "provided" vs "internal" Traefik, which could be again confusion-prone)

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
...
+
+```text
+          +-----------+    +-----------+    +---------+
+Client -> |  HAProxy  | -> |  Traefik  | -> |  REANA  |
+          +-----------+    +-----------+    +---------+
+```
+
+First of all, this is how you would set `X-Forwarded-Proto` when using HAProxy:
```
tiborsimko, line 122: Such a setup would require to set `X-Forwarded-Proto` in HAProxy as follows:

docs/administration/configuration/configuring-external-access/index.md
```diff
@@ -0,0 +1,136 @@
+# Configuring external access
```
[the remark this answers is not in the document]
mdonadoni: I also did not like much "external access", but I couldn't come up with a better term. I like _Configuring cluster ingress_!

docs/administration/configuration/configuring-external-access/index.md
mdonadoni: Yes, I also prefer _external traffic_ to _external access_

docs/administration/configuration/configuring-external-access/index.md
mdonadoni: In this context with "provided" I was referring to the "internal" Traefik that REANA deploys. The general idea also works for a "separate"/external instance of Traefik, but the configuration details might change a little bit depending on how that instance is deployed and configured. 

Should I perhaps remove "provided"?

_Let us now consider another setup in which HAProxy sits in front of Traefik: ..._

tiborsimko: If we use "provided", then we should probably list it under the "REANA" box, not as a separate box.

I see basically two solutions:

(1) We could change the use case to the one using "Manual ingress" and we say that REANA has internal ingress disabled and we put an extra Traefik instance before REANA; then the picture would be OK.

(2) We could keep the internal Traefik, but expose internal vs external Traefik situation clearer by painting a picture like:

```
           +--REANA---------------------+
           |            +----------+    |
Client --> | traefik -> | r-server |... |
           | enabled    +----------+    |
           +----------------------------+
```

This might perhaps overcomplicate things a bit, since we are starting to expose some of the internal components, but it would be closer to reality. And it would easily allow to explain also the other internal vs external traefik ingress use cases.

E.g. we could draw similar ASCII arts for all the sections above, showing this one for "Configuring default ingresses", whilst showing another one for "Setting up manual ingresses":

```
           +--External Ingress--+     +--REANA---------------------+
           |      traefik       |     |            +----------+    |
Client --> |       nginx        | --> | traefik -> | r-server |... |
           |        etc         |     | disabled   +----------+    |
           +--------------------+     +----------------------------+
```

Note also that people may be interested in how to use Nginx as ingress -- see https://github.com/reanahub/docs.reana.io/pull/148 -- which we may want to integrate into this page after merge. So it would be good to think about page structure accommodating nginx information in the near future.

OT1H, pictures of this kind would expose only some REANA internals such as r-ui or r-server, but that OTOH might be acceptable in a documentation page dedicated to ingresses. And we could expose r-w-controller, r-j-controller and friends in another page dedicated to the internal architecture.

If we go this way, then the ingress page ASCII art would be somewhat "deeper" than it currently is, but would still basically only "dip into" some of the REANA internals a bit, not bothering with anything which is not ingress related. (This would be exposed in another documentation page.) We could actually use dia/drawio/googleslides/etc to create these images instead of using ASCII art, so that they would have similar style in the "full picture" pages too.

(3) BTW, with the above in mind, the current nomenclature "Configuring default ingresses" vs "Setting up manual ingresses" might be perhaps better said "Configuring internal ingresses" vs "Configuring external ingresses"? Here I mean internal/external with respect to REANA, although people could think of "internal/external to cluster", which could be perhaps confusion-prone too.

Just some raw brain dump, we can chat IRL about pros/cons?

mdonadoni: I have pushed a new commit to track changes more easily, I will squash all of them when the PR is ready to be merged. I have also added one missing part for setting up manual ingresses.

docs/administration/configuration/configuring-cluster-ingress/index.md
```diff
@@ -0,0 +1,128 @@
...
+In these situations, you can make use of a non-standard `X-Forwarded-*` family of HTTP headers to preserve the details about the original connection.
+To fix the particular issue presented before, it is important to configure your proxy to provide correctly the `X-Forwarded-Proto` header.
+
+!!! warning
+
+    Using `X-Forwarded-*` headers can create security issues.
+    You must make sure that the `X-Forwarded-*` headers provided by the client are either discarded or overwritten by a proxy you trust.
+    See for example the [security concerns for X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#security_and_privacy_concerns).
+
+First of all, the provided Traefik instance must be configured to trust the `X-Forwarded-*` headers coming from HAProxy, so that they won't be overwritten:
```
tiborsimko, line 114: Since the previous text speaks about HAProxy mostly, I would inverse the order and first present what HAProxy configuration is needed, and only then speak of REANA's default proxy amendments.

(In this way the reader basically follows the ASCII art graph from left to right.)

This would give a text like:

You therefore need to configure HAPRoxy to set the `X-Forwarded-Proto` header for each incoming request:

... screenshot

and then configure REANA's provided Traefik ingress to trust the `X-Forwarded-*` headers coming from HAProxy:

... screenshot

docs/administration/configuration/configuring-external-access/index.md
tiborsimko: Good idea pushing a new commit and then squashing everything before merging :+1: The proxy text reads better now, I have left some cosmetic comments and mostly a section reordering comment.

docs/administration/configuration/configuring-cluster-ingress/index.md
tiborsimko: BTW we could also move the "Warning" box at the end of the main text,  in which case the new section ordering will stand out even clearer.  

We could also place the Warning box simply into the text and instead of "by a proxy you trust" we can directly say "by HAProxy". (And perhaps provide example in the screenshot?)

docs/administration/configuration/configuring-cluster-ingress/index.md
```diff
@@ -0,0 +1,128 @@
...
+                number: 80
+      host: <hostname>
+```
+
+## Using reverse proxy
+
+You may be interested in placing a (reverse) proxy service in front of REANA.
+This requires additional configuration to make REANA work correctly.
+
+Let us consider a setup in which HAProxy proxy handles connections to REANA:
```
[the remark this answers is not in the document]
mdonadoni: Good catch, it was not supposed to be like that!

docs/administration/configuration/configuring-cluster-ingress/index.md
mdonadoni: I have kept the warning box as it was before, as I think that it would lose visibility if we move it to the end or if we place it inside the text.

I have also kept the generic _a proxy you trust_ as I wanted this warning to be quite general, as it applies to every proxy (and also to all the applications, not only to REANA). As far as REANA is concerned, only the header `X-Forwarded-Proto` is considered (see [source](https://github.com/reanahub/reana-server/blob/f3bc6327d0580fd4b002211bfdbc97b72e0fc25b/reana_server/config.py#L206)). However, there is no way to tell Traefik to only trust that header, so it would probably be a good idea to overwrite/drop all the other `X-Forwarded-*`  headers, but this might not be needed depending on how the rest of the system is configured...

tiborsimko (approved):

PR #168, opened 2023-06-08
opening-sessions: custom jupyter image requirements

### LINKED ISSUE #163: jupyter: improve documentation regarding custom docker images
REANA supports running custom docker images instead of the default one ([`JUPYTER_INTERACTIVE_SESSION_DEFAULT_IMAGE`](https://github.com/reanahub/reana-workflow-controller/blob/23b2002263ae2fa729287c4a302e6c020b4d5dfb/reana_workflow_controller/config.py#LL127C1-L127C42)) when opening interactive notebooks.

We should improve the documentation by indicating which are the requirements that custom docker images should meet in order to run correctly (e.g. which port, which protocol, etc.)

giuseppe-steduto:
Closes #163


docs/running-notebooks/opening-sessions/index.md
```diff
@@ -25,6 +25,19 @@ https://reana.cern.ch/f8be55e4-5d18-43f9-b977-f773fdcab163?token=<your-reana-tok
...
 ```
 
+!!! warning
+    In order for the custom notebook image you supply to seamlessly integrate
+    with REANA, it must:
+
+    - have Jupyter installed
+    - expose port `8888`
+    - use the `HTTPS` protocol
+    - provide the [`start-notebook.sh`](https://github.com/jupyter/docker-stacks/blob/main/base-notebook/start-notebook.sh) script
```
giuseppe-steduto, line 35: I'm not sure whether it is better to mention this script and then recommend below to use an image based on one of the official Jupyter images, or to directly say that, as a requisite, the custom image should be based upon an official Jupyter one. What do you think?

docs/running-notebooks/opening-sessions/index.md
```diff
@@ -14,6 +14,12 @@ It could take several minutes to start the interactive session.
 Please note that it will be automatically closed after 7 days of inactivity.
 ```
 
+By clicking on the link you will have access to a Jupyter notebook interface:
+
+![jupyter-notebook](../../images/interactive-session-jupyter-notebook.png){.screenshot-browser-mockup}
+
+Keep in mind that starting the notebook might take a minute, if you see "Service unavailable", just wait and retry.
```
tiborsimko, line 21: After seeing the updated version, I would delete the "Keep in mind" phrase, because it disturbs the reading flow.  We are already warning about this in the console output shown above, so it should not be necessary  to add it also here!

docs/running-notebooks/opening-sessions/index.md
```diff
@@ -14,6 +14,12 @@ It could take several minutes to start the interactive session.
...
+
 !!! warning
```
tiborsimko, line 23: The Warning box also disrupts the reading flow. I would do put it at the end of the CLI chapter, i.e. right before the "Open from web interface" section. And I would add "Please note that it will be automatically closed after 7 days of inactivity" to the last example (using custom image) as well. WDYT?

tiborsimko (approved):

PR #165, opened 2023-06-05
notebooks: document new auto-closure feature

giuseppe-steduto:
- Adds a new documentation page for administrators on how to configure notebook auto-closure
- Improves the `running-notebooks` page for users, documenting how they are alerted about auto-closure of their sessions

Closes #154

PR #169, opened 2023-06-15
administration: add configuration of private registries

### LINKED ISSUE #159: docs: configuration of private docker registry
Add missing (administration) docs on how to configure a private docker registry.

In particular (non-exhaustive list!):
- Needed k8s secrets
- Helm values (e.g. `components.reana_workflow_controller.environment.IMAGE_PULL_SECRETS`)
- For GitLab private registry: how to create "service" account on GitLab

mdonadoni:
Closes #159

### DIFF
--- a/docs/administration/configuration/configuring-private-docker-registries/index.md
+++ b/docs/administration/configuration/configuring-private-docker-registries/index.md
@@ -1,0 +1,49 @@
+# Configuring private Docker registries
+
+REANA is able to execute workflows which make use of Docker images hosted on private Docker image registries.
+You will just need to save the necessary credentials in Kubernetes secrets, and then instruct REANA to pass these secrets to each job's pod.
+
+## Obtaining the credentials
+
+First of all, you will need to obtain the authentication credentials, and how to do so varies depending on which registry you are using.
+One possibility would be to create a special service account to access the registry.
+As an example, the [REANA service account](https://gitlab.cern.ch/reana) is used to authenticate to the private registry provided by CERN's GitLab instance, as shown in [Private Docker registries](../../../advanced-usage/access-control/private-docker-registries/#cern-gitlab).
+
+## Configuring Kubernetes and REANA
+
+After obtaining the credentials, you will have to save them in a Kubernetes secret, as explained in the related Kubernetes documentation page [Pull an Image from a Private Registry](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).
+In particular, this is how you would create the secret from an already existing Docker configuration file, after running `docker login`:
+
+```{ .console .copy-to-clipboard }
+$ kubectl create secret generic my-cern-gitlab-secret \
+    --from-file=".dockerconfigjson=$HOME/.docker/config.json" \
+    --type="kubernetes.io/dockerconfigjson"
+```
+
+As an alternative, you can create the secret by providing the username and password needed to authenticate to the private registry:
+
+```{ .console .copy-to-clipboard }
+$ kubectl create secret docker-registry my-cern-gitlab-secret \
+    --docker-server="https://gitlab-registry.cern.ch" \
+    --docker-username="johndoe" \
+    --docker-password="mysecretpassword" \
+    --docker-email="john.doe@example.org"
+```
+
+You should then set the [`components.reana_workflow_controller.environment.IMAGE_PULL_SECRETS`](https://github.com/reanahub/reana/tree/master/helm/reana) Helm value to the name of the secret, which in the above examples is `my-cern-gitlab-secret`:
+
+```yaml
+components:
+  reana_workflow_controller:
+    environment:
+      IMAGE_PULL_SECRETS: "my-cern-gitlab-secret"
+```
+
+If you would like to configure multiple private Docker registries, you should create one secret for each of them, for example `my-cern-gitlab-secret` for CERN GitLab and `my-ghcr-secret` for GitHub Container Registry. Then, you have to specify all the secrets in the `components.reana_workflow_controller.environment.IMAGE_PULL_SECRETS` Helm value, separating the names with commas:
+
+```yaml
+components:
+  reana_workflow_controller:
+    environment:
+      IMAGE_PULL_SECRETS: "my-cern-gitlab-secret,my-ghcr-secret"
+```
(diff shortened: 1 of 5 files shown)


docs/administration/configuration/configuring-private-docker-registries/index.md
```diff
@@ -0,0 +1,49 @@
+# Configuring private docker registries
```
giuseppe-steduto, line 1:
```suggestion
# Configuring private Docker registries
```

docs/administration/configuration/configuring-private-docker-registries/index.md
```diff
@@ -0,0 +1,49 @@
...
+You should then set the [`components.reana_workflow_controller.environment.IMAGE_PULL_SECRETS`](https://github.com/reanahub/reana/tree/master/helm/reana) Helm value to the name of the secret, which in the above examples is `<your-credentials-secret>`:
+
+```yaml
+components:
+  reana_workflow_controller:
+    environment:
+      IMAGE_PULL_SECRETS: "<your-credentials-secret>"
+```
+
+If you would like to configure multiple private docker registries, you should create one secret for each of them. Then, you have to specify all the secrets in the `components.reana_workflow_controller.environment.IMAGE_PULL_SECRETS` Helm value, separating the names with commas:
```
giuseppe-steduto, line 42:
```suggestion
If you would like to configure multiple private Docker registries, you should create one secret for each of them. Then, you have to specify all the secrets in the `components.reana_workflow_controller.environment.IMAGE_PULL_SECRETS` Helm value, separating the names with commas:
```

docs/administration/configuration/configuring-private-docker-registries/index.md
```diff
@@ -0,0 +1,49 @@
...
+
+```yaml
+components:
+  reana_workflow_controller:
+    environment:
+      IMAGE_PULL_SECRETS: "<your-first-secret>,<your-second-secret>"
```
giuseppe-steduto, line 48:
```suggestion
      IMAGE_PULL_SECRETS: "<your-first-secret-name>,<your-second-secret-name>"
```

What do you think? Adding "name" might avoid that the users think they should type the secret itself rather than the name with which they added it to Kubernetes. The same goes for the example above with `<your-credentials-secret>`

docs/administration/configuration/configuring-private-docker-registries/index.md
```diff
@@ -0,0 +1,49 @@
...
+
+As an alternative, you can create the secret by providing the username and password needed to authenticate to the private registry:
+
+```{ .console .copy-to-clipboard }
+$ kubectl create secret docker-registry <your-credentials-secret> \
+    --docker-server=<your-registry-server> \
+    --docker-username=<your-name> \
+    --docker-password=<your-password> \
+    --docker-email=<your-email>
+```
```
giuseppe-steduto, line 31: In [another discussion](https://github.com/reanahub/docs.reana.io/pull/165#discussion_r1226590136), we considered the idea of using the `!!! tip` admonition for `kubectl`-related tips and how-tos throughout the admin-related documentation. I think it could fit well here, do you agree?

docs/administration/configuration/configuring-private-docker-registries/index.md
```diff
@@ -0,0 +1,49 @@
+# Configuring private docker registries
+
+REANA is able to execute workflows which make use of Docker images hosted on private Docker image registries.
+You will just need to save the necessary credentials in Kubernetes secrets, and then instruct REANA to pass these secrets to each job's pod.
+
+## Obtaining the credentials
+
+First of all, you will need to obtain the authentication credentials, and how to do so varies depending on which registry you are using.
+One possibility would be to create a special service account to access the registry.
+As an example, the [REANA account](https://gitlab.cern.ch/reana) is used to authenticate to the private registry provided by CERN's GitLab instance, as shown in [Private Docker registries](../../../advanced-usage/access-control/private-docker-registries/#cern-gitlab).
```
tiborsimko, line 10: Perhaps we can call it the "REANA service account" user, here and in the linked user-facing documentation.

docs/administration/configuration/configuring-private-docker-registries/index.md
tiborsimko: Personally I don't like the use of `<variable>` in shell command examples, because it may be confusing WRT shell redirections.

Better to use `john.doe@example.org` , `johndoe`  etc so that readers will see a "realistic" example.

(Ditto `<path/to/.docker/config.json>` can better be `$HOME/.docker/config.json` for example.)

docs/administration/configuration/configuring-private-docker-registries/index.md
tiborsimko: I would call them `mydockerhubsecret`, `myghrcsecret`, etc. explaining that this secret is for Docker Hub, that secret for GHCR.  (Just an example.)

Basically the idea is to create a scenario and provide concretely usable example that can be also "copy-pasted" and amended, instead of providing just a theoretical description. Hence the use of `john.doe@example.org` etc.

(See also the comment about `<foo>` syntax above.)

mdonadoni: I have updated the examples to refer to the CERN GitLab registry, but if you prefer to use a more known registry I can change to DockerHub

docs/administration/configuration/configuring-private-docker-registries/index.md
mdonadoni: > In [another discussion](https://github.com/reanahub/docs.reana.io/pull/165#discussion_r1226590136), we considered the idea of using the `!!! tip` admonition for `kubectl`-related tips and how-tos throughout the admin-related documentation. I think it could fit well here, do you agree?

Given that this is a necessary step to reach the end of the guide and this isn't an "additional tip" as in the other case, I think it's better to keep these commands as part of the main text instead of using `!!! tip` boxes.

docs/administration/configuration/index.md
```diff
@@ -3,6 +3,7 @@
 - [Configuring compute backends](configuring-compute-backends)
 - [Configuring interactive sessions](configuring-interactive-sessions)
 - [Configuring storage volumes](configuring-storage-volumes)
+- [Configuring private docker registries](configuring-private-docker-registries)
```
giuseppe-steduto, line 6:
```suggestion
- [Configuring private Docker registries](configuring-private-docker-registries)
```

giuseppe-steduto (commented):
Looks good! There is only one capital letter missing

giuseppe-steduto (approved):

PR #180, opened 2023-09-12
adding instructions to connect to custom Rucio instance

egazzarr:


tiborsimko (approved):
Thanks! I have had a similar local branch with an upgrade of reana-auth-rucio sidecar image version so I have merged the two :wink:

PR #185, opened 2023-10-20
user-quotas: update image of workflow deletion modal

mdonadoni:
Closes reanahub/reana-ui#349


giuseppe-steduto (commented):
Good! However, now we have two different styles of images in the same page: the old-style screenshot from Firefox, and the new version, with the CSS class. How about changing the other images so that all the screenshots look the same?

PR #187, opened 2023-11-08
launching-workflows: mention launcher badge generator

giuseppe-steduto:
Closes reanahub/reana-ui#324.


docs/running-workflows/launching-workflows/index.md
```diff
@@ -135,22 +135,13 @@ The standard set of badges look as follows:
 
 ![Launch with Snakemake REANA at CERN](https://www.reana.io/static/img/badges/launch-with-snakemake-on-reana-at-cern.svg)
 
-If you would like to generate your own custom badge, you can use external services such as [shields.io](https://shields.io) or [badgen.net](https://badgen.net).
```
mdonadoni, line 138: I think we should keep the part about generating custom badge designs using external services, as the badge creator is not a substitute for it (e.g. there is no way to change the text inside the badge, etc.)

docs/running-workflows/launching-workflows/index.md
```diff
@@ -135,22 +135,13 @@ The standard set of badges look as follows:
...
+To streamline the badge creation process, you can use the tool accessible on REANA at the
```
mdonadoni, line 138: What if we keep the sections about Markdown and HTML, and add a new one before those for the badge creator? This way we can show both the "automatic" way to generate badges, and the "manual" way to do it with Markdown and HTML.

What do you think?

mdonadoni, line 138: Maybe let's explain that this is instance-dependent?
Something like: ... accessible on your REANA instance at `/launcher-badge`. If you have a CERN account, you can make use of launch badge creator of the available REANA@CERN instance.

docs/running-workflows/launching-workflows/index.md
```diff
@@ -135,22 +135,13 @@ The standard set of badges look as follows:
...
-[![Launch on REANA badge](https://www.reana.io/static/img/badges/launch-on-reana.svg)](https://reana.cern.ch/launch?url=https%3A%2F%2Fgithub.com%2Freanahub%2Freana-demo-root6-roofit)
-```
-
-This example points to the RooFit analysis demo. You should replace the `url` to point to your own analysis sources.
-
-### HTML
-
-You can include one of the above badges into your HTML documentation pages using the following syntax:
+Upon visiting the launcher badge page, you will be able to generate the necessary markdown syntax to include the badge
+in your documentation. This tool also provides a generic URL for initiating the workflow, which can be integrated into
```
mdonadoni, line 142: Something like:

```suggestion
in your documentation. This tool also provides you the URL for executing your workflow, which can be integrated into
```

To emphasize that the tool generates the URL specific to a workflow, and not a generic one that needs to be customised. What do you think?

docs/running-workflows/launching-workflows/index.md
giuseppe-steduto: Okay, I agree! Should we do something similar for the other examples in the rest of the page? Even though maybe for them it's probably better to leave `reana.cern.ch`...

PR #189, opened 2023-11-14
docs: add share-status command

DaanRosendal:
Adds a new command to the docs for retrieving whom a workflow is shared with.

Closes reanahub/reana-client#686


mdonadoni:
Included in https://github.com/reanahub/docs.reana.io/pull/190

PR #191, opened 2023-12-05
refresh CWL v1.0 support results

mr-c:
All the extra failing tests were due to timeouts (10 minutes)...


giuseppe-steduto:
Thank you very much for opening the PR! We have rerun the tests on the latest REANA version and found that some timeout failing tests were fixed, so we opened another PR: #196.

PR #203, opened 2024-04-05
docs(ingress): avoid deprecated annotation to set ingress class (#203)

mdonadoni:
Closes reanahub/reana#742


docs/administration/configuration/configuring-cluster-ingress/index.md
```diff
@@ -10,6 +10,7 @@ REANA creates a single Ingress object by default, which can be customized with t
 
 - `ingress.enabled` enables or disables the creation of Ingresses
 - `ingress.annotations` is used to define which custom annotations should be applied to the Ingress resource
+- `ingress.ingress_class_name` is used to select the desired [Ingress class](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class), and it can be set to `null` to choose the default Ingress class of the cluster
```
tiborsimko, line 13: ... "and can be set to `null` to use the default Ingress class of the cluster".

PR #212, opened 2024-11-29
docs(dask): add Dask docs for users and admins

Alputer:
Adds documentation for REANA's Dask support, covering both the user-facing usage guide (`docs/advanced-usage/dask/`) and the administrator configuration guide (`docs/administration/configuration/configuring-dask/`).

The user guide explains how to request a dedicated Dask cluster via `reana.yaml` (`image`, `number_of_workers`, `single_worker_memory`, `single_worker_threads`), query cluster limits with `reana-client info`, and use the Dask dashboard. The admin guide documents the relevant Helm values for enabling Dask and configuring autoscaling, cluster memory, and per-worker worker/memory/thread defaults and limits.


docs/administration/configuration/configuring-dask/index.md
```diff
@@ -0,0 +1,98 @@
...
+
+This value defines the upper limit on the number of workers a user can request in their reana.yaml, even if their workflow does not reach the cluster_memory_limit. It acts as a safeguard to prevent users from requesting an excessive number of workers with very low memory allocations (e.g., 100 workers with only 30Mi memory).
+
+```yaml
+dask:
+  cluster_max_number_of_workers: 50
+```
+
+In this case, users can request up to 50 workers in their reana.yaml. If they attempt to request more than the maximum limit, the system will cap the cluster size to 50 workers, regardless of the memory limits.
+
```
michaelbuchar: I think the max-worker behavior is described incorrectly here. This section says REANA will cap an oversized request to the configured maximum, but the actual server-side validation rejects the workflow submission when number_of_workers exceeds REANA_DASK_CLUSTER_MAX_NUMBER_OF_WORKERS. That also contradicts the user facing Dask page in this same PR, which already shows a validation error for this case. I think this paragraph should say the request is rejected rather than capped.

docs/advanced-usage/dask/index.md
```diff
@@ -0,0 +1,88 @@
...
+An example configuration:
+
+```yaml
+...
+resources:
+   dask:
+      image: docker.io/coffeateam/coffea-dask-cc7:0.7.22-py3.10-g7f049
+      number_of_workers: 5
+      single_worker_memory: 8Gi
+...
```
michaelbuchar: This example is a bit ambiguous about where the Dask block belongs in reana.yaml. As written, it shows a bare resources: section, but the schema expects this under workflow.resources.dask. I think it would be clearer to show a minimal complete example here so users can copy it without guessing the nesting.

docs/advanced-usage/dask/index.md
```diff
@@ -0,0 +1,88 @@
+# Dask
+
+REANA supports the integration of Dask clusters to provide scalable, distributed computing capabilities for workflows. This documentation explains how to set up and configure a dedicated Dask cluster for your workflow, query cluster settings, and utilize Dask's features such as the dashboard for monitoring your workflows.
+
+## Setting up Dask cluster
+
+To configure your Dask cluster, you can set three different variables in `reana.yaml`
```
michaelbuchar: I think single_worker_threads should be documented as well. The current docs describe image, number_of_workers, and single_worker_memory, but single_worker_threads is also part of the supported Dask workflow schema and has corresponding cluster defaults/limits. As a result, the user and admin docs currently read as if threads are not configurable, which is incomplete for anyone trying to tune Dask worker resources.

docs/administration/configuration/configuring-dask/index.md
```diff
@@ -0,0 +1,98 @@
...
+Dask support is disabled by default in REANA. If you would like to enable it so that users can ask for a Dask cluster for their workflow, you can set [`dask.enabled`](https://github.com/reanahub/reana/tree/master/helm/reana) Helm value to `true`.
+
+!!! warning
+
+    When upgrading your cluster using the `helm upgrade` command instead of `helm install`, Dask custom resource definitions may not install properly due to how Helm manages CRDs.
+    When using `helm upgrade`, you'll need to manually install Dask CRDs:
+
+    ```console
+    $ helm install --repo https://helm.dask.org --generate-name dask-kubernetes-operator
+    ```
```
michaelbuchar, line 18: The CRD warning is worded imprecisely. This says “manually install Dask CRDs”, but the suggested helm install ... dask-kubernetes-operator command installs the operator chart, not just the CRDs. Since REANA already declares that chart as a dependency in [Chart.yaml (line 47)](/Users/michaelbuchar/Desktop/reana/project/reana/src/reana/helm/reana/Chart.yaml:47), this should be clarified to avoid duplicate install confusion.

docs/advanced-usage/dask/index.md
CameronMcClymont: Yes you're right, thanks. This must've been added at some point since the PR was opened cause it's relatively old. Adding now.

docs/advanced-usage/dask/index.md
CameronMcClymont: Expanded this to a more complete example now to make it more clear.

docs/administration/configuration/configuring-dask/index.md
CameronMcClymont: Reworded this line to improve clarity.

docs/administration/configuration/configuring-dask/index.md
CameronMcClymont: Yeah I think that's what he meant by 'capped', but updated to use 'rejected' now to improve clarity here too.

docs/advanced-usage/dask/index.md
```diff
@@ -0,0 +1,106 @@
+# Dask
+
+REANA supports the integration of Dask clusters to provide scalable, distributed computing capabilities for workflows. This documentation explains how to set up and configure a dedicated Dask cluster for your workflow, query cluster settings, and utilize Dask's features such as the dashboard for monitoring your workflows.
+
+## Setting up Dask cluster
+
+To configure your Dask cluster, you can set four different variables in `reana.yaml`
+
+1. **image** (mandatory)
+   Specifies the Docker image to be used by the Dask workers, scheduler, and the job pod that executes your analysis. Ensure the image includes all dependencies required for your workflow.
```
michaelbuchar: The user guide describes workflow.resources.dask.image too broadly. [docs/advanced-usage/dask/index.md#L9-L10](https://github.com/reanahub/docs.reana.io/blob/fb995ccb37672e443d99a581eea23ad0e77df415/docs/advanced-usage/dask/index.md#L9-L10) says this image is used by the Dask workers, scheduler, and the job pod that executes the analysis, but the schema and controller only use it for the Dask scheduler and workers. The workflow job itself still uses the step environment image. See [reana_analysis_schema.json (line 122)](/Users/michaelbuchar/Desktop/reana/project/reana/src/reana-commons/reana_commons/validation/schemas/reana_analysis_schema.json:122) and [dask.py (line 186)](/Users/michaelbuchar/Desktop/reana/project/reana/src/reana-workflow-controller/reana_workflow_controller/dask.py:186). As written, this can mislead users into thinking the Dask cluster image also controls the analysis container.

CameronMcClymont: Nice catch, updated to clarify.

michaelbuchar: Lg!

michaelbuchar (approved):

docs/advanced-usage/dask/index.md
```diff
@@ -0,0 +1,106 @@
...
+The `dask` block lives under `workflow.resources` in your `reana.yaml`. A minimal example looks as follows:
+
+```yaml
+workflow:
+  type: serial
+  resources:
+    dask:
+      image: docker.io/coffeateam/coffea-dask-cc7:0.7.22-py3.10-g7f049
+      number_of_workers: 5
+      single_worker_memory: 8Gi
```
tiborsimko: **🤖 D212-01: Combined cluster-memory limit is undocumented — and the page's own example violates it (MEDIUM)**

The "Setting up Dask cluster" section documents three of the four resource-limit checks a Dask workflow submission goes through, but the server enforces a fourth one: the *combined* cluster memory, i.e. `number_of_workers × single_worker_memory`, must not exceed the administrator's `dask.cluster_max_memory_limit` (see `validate_dask_limits` in reana-server's `validation.py`):

```console
$ reana-client run -w my-dask-workflow
...
==> ERROR: Cannot create workflow:
The "memory" requested in the dask resources exceeds the limit (16Gi).
Decrease the number of workers requested or amount of memory consumed by a single worker.
```

Concretely, the minimal example on this very page requests `number_of_workers: 5` with `single_worker_memory: 8Gi`, i.e. 40Gi combined — 2.5× the default `dask.cluster_max_memory_limit` of 16Gi. Every individually documented limit is respected (5 ≤ 20 workers, 8Gi ≤ 8Gi per worker, 4 ≤ 8 threads), yet copy-pasting the example onto a default-configured cluster is rejected with the error above, which the page gives readers no way to understand.

The check is also easy to trip without setting any memory at all, because defaults are substituted before the multiplication: on a default cluster, `number_of_workers: 10` alone is rejected (10 × default 2Gi = 20Gi > 16Gi) despite being well within the 20-worker maximum. The corresponding `reana-client info` line ("The maximum memory limit for Dask clusters created by users") appears in the example output below, but its interaction with `number_of_workers` and `single_worker_memory` is never explained. (The admin page does explain it, in "Limiting Cluster Memory".)

Suggested fix — two parts:

1. Change the example to fit within default limits, e.g. `single_worker_memory: 2Gi` (5 × 2Gi = 10Gi ≤ 16Gi).
2. Add a short note after the four field descriptions (before "The `dask` block lives under `workflow.resources`…"), e.g.:

> Note also that the *total* memory of the requested cluster, i.e. `number_of_workers` multiplied by `single_worker_memory` (using the cluster defaults for any value you do not set), must fit within the maximum Dask cluster memory limit configured by your REANA administrator. Otherwise the workflow submission is rejected with:
>
> ```console
> ==> ERROR: Cannot create workflow:
> The "memory" requested in the dask resources exceeds the limit (16Gi).
> Decrease the number of workers requested or amount of memory consumed by a single worker.
> ```

docs/advanced-usage/dask/index.md
```diff
@@ -0,0 +1,106 @@
+# Dask
+
+REANA supports the integration of Dask clusters to provide scalable, distributed computing capabilities for workflows. This documentation explains how to set up and configure a dedicated Dask cluster for your workflow, query cluster settings, and utilize Dask's features such as the dashboard for monitoring your workflows.
```
tiborsimko: **🤖 D212-02: User page does not mention that Dask support is cluster-dependent (LOW)**

Dask support is disabled by default (`dask.enabled: false` in the Helm chart), so on deployments retaining the default the very first error a reader of this page will hit is:

```console
$ reana-client run -w my-dask-workflow
...
==> ERROR: Cannot create workflow:
Dask workflows are not allowed in this cluster.
```

(raised by `validate_dask_limits` in reana-server's `validation.py` whenever a `dask:` block is present and the cluster has Dask disabled). The page currently presents Dask as universally available and never mentions this error, even though the "Querying Dask Settings and Limits" example output already contains the relevant flag ("Dask workflows allowed in the cluster: True") without drawing attention to it.

Suggested fix — add one sentence to the introduction (or at the start of "Setting up Dask cluster"), e.g.: "Note that Dask support is optional and may not be enabled on your REANA cluster; you can check the `Dask workflows allowed in the cluster` line in the `reana-client info` output described below. If it is disabled, workflow submissions containing a `dask:` block are rejected with the error `Dask workflows are not allowed in this cluster.`"

docs/administration/configuration/configuring-dask/index.md
```diff
@@ -0,0 +1,132 @@
+# Configuring Dask
+
+Dask integration in REANA allows users to request dedicated Dask clusters for their workflow requirements. Each cluster operates independently, providing the computational resources necessary to efficiently execute workflows.
+
+## Enabling Dask
+
+Dask support is disabled by default in REANA. If you would like to enable it so that users can ask for a Dask cluster for their workflow, you can set [`dask.enabled`](https://github.com/reanahub/reana/tree/master/helm/reana) Helm value to `true`.
```
tiborsimko, line 7: **🤖 D212-03: Dask dashboard availability depends on Traefik, which neither page mentions (LOW)**

reana-workflow-controller only creates the Dask dashboard ingress when Traefik is available — `create_dask_resources` in `reana_workflow_controller/dask.py` guards it with `if TRAEFIK_ENABLED or TRAEFIK_EXTERNAL`, which map to the `traefik.enabled` and `traefik.external` Helm values. On a cluster deployed with `traefik.enabled: false` and no external Traefik (e.g. sites using a different ingress controller), Dask clusters work but the dashboard documented in the user page's "Dask Dashboard" section is never exposed. Worse, the web UI still shows the "Dashboard" badge in that case — `WorkflowBadges.js` renders it whenever the workflow has a running Dask service, with no knowledge of whether the ingress exists — so users on such deployments get a dead link.

Since this admin page is where a site administrator would look to understand what enabling Dask entails, it is worth one sentence here so admins are not surprised when users report a broken dashboard link.

Suggested fix — add a note to the "Enabling Dask" section, e.g.: "Note that the Dask dashboard of each cluster is exposed via Traefik, so it is only available when your deployment uses Traefik (`traefik.enabled` or `traefik.external`). Dask workflows still run fine without Traefik, but the dashboard links will not be served." Optionally mirror one sentence in the user page's "Dask Dashboard" section ("depending on your REANA cluster's configuration, the dashboard may not be available — ask your administrator").

docs/administration/configuration/configuring-dask/index.md
```diff
@@ -0,0 +1,132 @@
...
+  cluster_default_number_of_workers: 3
+```
+
+In this case, every Dask cluster will start with 3 workers unless a different number is provided in the reana.yaml.
+
+If the cluster administrator does not override the `dask.cluster_default_number_of_workers` variable, it is set to `2` by default.
+
+### [`dask.cluster_max_number_of_workers`](https://github.com/reanahub/reana/tree/master/helm/reana)
+
+This value defines the upper limit on the number of workers a user can request in their reana.yaml, even if their workflow does not reach the cluster_memory_limit. It acts as a safeguard to prevent users from requesting an excessive number of workers with very low memory allocations (e.g., 100 workers with only 30Mi memory).
```
tiborsimko: **🤖 D212-04: `cluster_memory_limit` is not the name of the Helm value (NIT)**

> This value defines the upper limit on the number of workers a user can request in their reana.yaml, even if their workflow does not reach the cluster_memory_limit.

There is no `cluster_memory_limit` Helm value; the value documented earlier on this very page is `dask.cluster_max_memory_limit`. Using a second spelling for the same knob makes it harder to grep/cross-reference.

Suggested fix — "…even if their workflow does not reach the `dask.cluster_max_memory_limit`."

docs/administration/configuration/configuring-dask/index.md
```diff
@@ -0,0 +1,132 @@
...
+
+Dask support is disabled by default in REANA. If you would like to enable it so that users can ask for a Dask cluster for their workflow, you can set [`dask.enabled`](https://github.com/reanahub/reana/tree/master/helm/reana) Helm value to `true`.
+
+!!! warning
+
+    REANA declares the Dask Kubernetes operator (including its custom resource definitions) as a Helm chart dependency, so a fresh `helm install` of REANA sets everything up automatically. However, when updating an existing cluster with `helm upgrade` instead of `helm install`, the Dask custom resource definitions may not be installed, because Helm does not manage CRDs on upgrades.
+    In that case, install the Dask Kubernetes operator chart manually, which also registers the missing CRDs:
+
+    ```console
+    $ helm install --repo https://helm.dask.org --generate-name dask-kubernetes-operator
```
tiborsimko: **🤖 D212-05: CRD remediation command installs a second, unpinned Dask operator (MEDIUM)**

> ```console
> $ helm install --repo https://helm.dask.org --generate-name dask-kubernetes-operator
> ```

This command does register the missing CRDs, but as a side effect it deploys a complete second operator. After a `helm upgrade` of REANA, the cluster already runs the operator that ships as REANA's chart dependency — pinned to `dask-kubernetes-operator` 2025.4.3 in `Chart.yaml` and running REANA's customised image (`reanahub/reana-dask-kubernetes-operator`, with REANA-specific database environment) — and is missing *only* the CRDs. Following the instruction as written leaves two operators watching the same `DaskCluster` resources cluster-wide: REANA's customised one and a vanilla upstream one at whatever the latest chart version happens to be (the command is unversioned). The two can race on reconciling the same clusters, the vanilla operator lacks the REANA-specific behaviour of the customised image, and an unpinned install can register CRDs from a newer upstream release than the pinned operator expects. The page never says to uninstall the extra release afterwards, so this state persists.

Suggested fix — instruct admins to install only the CRDs, at the chart version matching REANA's dependency pin:

```console
$ helm show crds --repo https://helm.dask.org --version <operator-chart-version> dask-kubernetes-operator | kubectl apply --server-side -f -
```

where `<operator-chart-version>` is the `dask-kubernetes-operator` dependency version in the `Chart.yaml` of the REANA chart release being deployed (`2025.4.3` at the time of writing). This is consistent with Helm's documented CRD lifecycle — CRDs are installed on `helm install` but never upgraded or deleted by Helm — which the warning already cites indirectly via the GitHub issue.

docs/administration/configuration/configuring-dask/index.md
```diff
@@ -0,0 +1,132 @@
...
+This value determines the default number of workers assigned to a Dask cluster if the user does not explicitly specify the `number_of_workers` field in their `reana.yaml` workflow configuration file. Setting this value ensures that all Dask clusters start with a reasonable number of workers to handle typical workloads without requiring user input.
+
+For example:
+
+```yaml
+dask:
+  cluster_default_number_of_workers: 3
+```
+
+In this case, every Dask cluster will start with 3 workers unless a different number is provided in the reana.yaml.
```
tiborsimko: **🤖 D212-06: `number_of_workers` semantics are wrong when the autoscaler is enabled (MEDIUM)**

> In this case, every Dask cluster will start with 3 workers unless a different number is provided in the reana.yaml.

This is only true when the autoscaler is disabled. In the default configuration (`dask.autoscaler_enabled: true`), reana-workflow-controller creates the `DaskCluster` with `worker.replicas = 0` and wires `number_of_workers` into the `DaskAutoscaler`'s `spec.maximum` (see `_prepare_cluster` and `_prepare_autoscaler` in `reana_workflow_controller/dask.py`). So out of the box a cluster *starts with zero workers* and scales up to at most 3; it never "starts with 3 workers". The static-worker-count reading only applies after an admin sets `dask.autoscaler_enabled: false`.

The "Configuring Autoscaler" section above actually describes the scale-up-to-N behaviour correctly, so as written the page contradicts itself. The user page has the same imprecision in milder form ("Defines the number of Dask workers for your cluster").

Suggested fix — reword to cover both modes, e.g.: "In this case, every Dask cluster can scale up to 3 workers unless a different number is provided in the reana.yaml (with the autoscaler disabled, the cluster starts with a fixed 3 workers instead)." On the user page, describe `number_of_workers` as "the (maximum) number of Dask workers for your cluster; with the autoscaler enabled — the default — the cluster scales between zero and this many workers depending on load".

docs/advanced-usage/dask/index.md
tiborsimko: Amended during merge.

docs/administration/configuration/configuring-dask/index.md
tiborsimko: Amended during merge.

docs/administration/configuration/configuring-dask/index.md
tiborsimko: Amended during merge.

docs/administration/configuration/configuring-dask/index.md
tiborsimko: Amended during merge.

docs/advanced-usage/dask/index.md
tiborsimko: Amended during merge.

docs/administration/configuration/configuring-dask/index.md
tiborsimko: Amended during merge.

tiborsimko (approved):

PR #217, opened 2025-02-28
docs(configuring-storage-volumes): fix shared volume mix-up (#217)

### LINKED ISSUE #216: Possible word mix up in Configuring storage volumes
At https://docs.reana.io/administration/configuration/configuring-storage-volumes/#configuration you write:

> Note that typically, the shared storage volume can be huge (even tens of terabytes), while the shared storage volume can be as small as a few gigabytes, depending also on whether you are using an external database or not.

I assume this should be

> Note that typically, the shared storage volume can be huge (even tens of terabytes), while the **infrastructure** volume can be as small as a few gigabytes, depending also on whether you are using an external database or not.

tiborsimko:
Closes #216


Alputer (approved):
LGTM!

PR #222, opened 2025-11-12
docs(administration): add quota management endpoints docs

### LINKED ISSUE #747: rest: add optional quota management endpoints
### Current behaviour

The CPU and Disk quota for users can be set through command-line by REANA administrators:

```
kubectl exec -i -t deployment/reana-server -c rest-api -- flask reana-admin quota-set -e john.doe@example.org -r disk -l 429496729600 --admin-access-token $REANA_ADMIN_ACCESS_TOKEN
```

This is perfectly fine, however for certain deployment scenarios, one would like to achieve this over REST API without having direct Kubernetes cluster access, so that first support line could manage user quotas without having any cluster privileges.

### Desired behaviour

It would be therefore good to introduce new REST API endpoints that would allow to consult and set the quota for users remotely.

This new REST API should be accessible only after providing certain secret which is not necessarily the same as REANA_ADMIN_ACCESS_TOKEN. This will allow to support scenarios where the frontend support line could be granted rights to manage user quotas without having access to much else.

This secret could be set in a Helm value. And, if the secret would be empty (default value), this functionality would be disabled (and so quota management would be only available from the command line, as is the case now).

The calling of the new REST API endpoints could be simply via `curl` commands. (We don't need to enrich reana-client.) The admin documentation under `docs.reana.io` should then be updated accordingly.

CameronMcClymont:
Closes reanahub/reana-server#747

### DIFF
--- a/docs/administration/configuration/configuring-user-quotas/index.md
+++ b/docs/administration/configuration/configuring-user-quotas/index.md
@@ -25,7 +25,7 @@ Here is an example of a mixed configuration where, in addition to nightly CPU an
 ```yaml
 quota:
   enabled: true
-  # update quota usage everyday at 3AM
+  # update quota usage every day at 3AM
   periodic_update_policy: "0 3 * * *"
   # after workflow finishes or fails update its cpu usage
   workflow_termination_update_policy: "cpu"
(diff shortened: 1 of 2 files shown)


michaelbuchar (approved):
LGTM!

michaelbuchar:
Maybe we should add this PR to this issue as well: https://github.com/reanahub/reana-server/issues/747?

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -61,8 +63,61 @@ $ kubectl exec -i -t deployment/reana-server -- /bin/bash
...
+    environment:
+      REANA_QUOTA_MANAGEMENT_SECRET: "my_secret"
+```
+
+If the secret is unset or empty, quota management via the REST API is disabled.
+
+To get the current CPU quota limit and usage for the user `john.doe@example.org`, run:
+
+```console
+$ curl -v -H "X-Quota-Management-Secret: my_secret" \
```
tiborsimko: ### AI review

**Bug: Unquoted `&` in URL will break in shell.**

The `&` in the URL is not quoted, so the shell will interpret it as a background operator — `resource_type=cpu` will be treated as a separate command.

Suggested fix — quote the URL:

```console
$ curl -v -H "X-Quota-Management-Secret: my_secret" \
       "https://reana.cern.ch/api/quota?email=john.doe@example.org&resource_type=cpu"
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -61,8 +63,61 @@ $ kubectl exec -i -t deployment/reana-server -- /bin/bash
 You can now set a custom quota limit to selected users:
 
 ```console
-# flask reana-admin quota-set -e john.doe@example.org -r disk -l 250000 --admin-access-token $REANA_ACCESS_TOKEN
+# flask reana-admin quota-set -e john.doe@example.org -r disk -l 250000 --admin-access-token $REANA_ADMIN_ACCESS_TOKEN
```
tiborsimko: ### AI review

**Minor: `$REANA_ACCESS_TOKEN` renamed to `$REANA_ADMIN_ACCESS_TOKEN`.**

This looks intentional (clearer naming), but worth confirming the variable actually exists in the server pod environment, since this is modifying existing documentation.

tiborsimko:
### AI review

**Minor: CERN-specific URL in examples.**

The curl examples use `https://reana.cern.ch/api/quota`. If the docs are for the general REANA project, a generic placeholder like `https://reana.example.org/api/quota` might be more appropriate. Ignore if the intended audience is CERN users.

tiborsimko:
### AI review

**Curl examples should use `--fail-with-body` to return non-zero exit code on errors.**

Currently, the curl examples return exit code 0 even when the server responds with 4xx/5xx, which makes it hard to use them in scripts. Adding `--fail-with-body` (available since curl 7.76.0) makes curl return exit code 22 on HTTP errors while still printing the JSON error body:

```console
$ curl --fail-with-body -H "X-Quota-Management-Secret: mysecret" \
       -H "Content-Type: application/json" \
       -X POST \
       -d '{"email": "john.doe@cern.ch", "resource_type": "gpu", "limit": 10000000}' \
       https://localhost:30443/api/quota
{"message":"ERROR: Provided resource 'gpu' does not exist. Available resources: cpu, disk"}
curl: (22) The requested URL returned error: 400
```

Both the GET and POST curl examples in the documentation should include `--fail-with-body`.

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -61,8 +63,61 @@ $ kubectl exec -i -t deployment/reana-server -- /bin/bash
...
 Quota limit 250000 for 'disk (shared storage)' successfully set to users ('john.doe@example.org',).
 ```
 
 You can learn more about the `quota-set` administrative command options by running `flask reana-admin quota-set --help`.
+
+### Using the REST API
```
tiborsimko: For the beginning of the documentation, if we merge it into `master`, it'll be deployed on production docs.reana.io instance. Hence it'll be good to qualify that this feature comes with this-and-this REANA version. 

Also, we could explicitly speak about "Quota management REST API" as we called it elsewhere.

This would give something like:

```
### Using the Quota management REST API

As of REANA 0.95 release series, you can also set quota limits 
for specific users using the new Quota Management REST API at `/api/quota`.

....
```

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed, good catch!

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Should be fixed!

tiborsimko (approved):

PR #224, opened 2026-04-27
docs(administration): add quota period accounting docs

michaelbuchar:
Closes https://github.com/reanahub/reana-server/issues/774


tiborsimko (commented):
Inline review of the documentation changes. Only one inline item (MEDIUM); the docs are otherwise accurate against the implementations in reana-db #255 / reana-server #768 / reana #931.

One adjacent item that's outside this diff but worth a follow-up: the existing `GET /api/quota` JSON response example at lines 139-145 of `docs/administration/configuration/configuring-user-quotas/index.md` doesn't yet show `quota_period_months` and `quota_period_start_at`. Once reana-server #768 lands, GET also returns those fields, so the example will be incomplete. Either update it as part of this PR or open a follow-up.

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +163,70 @@ The response contains the new limit and the current usage for the given user and
...
+Quota period replenished successfully.
+```
+
+For service integrations, the same periodic CPU fields can be managed over the
+quota management REST API:
+
+```console
+$ curl -X PATCH https://reana.example.org/api/quota \
+    -H 'Content-Type: application/json' \
+    -H 'X-Quota-Management-Secret: $REANA_QUOTA_MANAGEMENT_SECRET' \
```
tiborsimko: 🤖 **Single quotes prevent `$REANA_QUOTA_MANAGEMENT_SECRET` from expanding (MEDIUM)**

The curl example wraps the header in single quotes, which means bash will not expand the variable — the literal string `$REANA_QUOTA_MANAGEMENT_SECRET` will be sent as the secret value, the server will compare it against the configured secret, and the request will fail with 403. A user copy-pasting the example as-is will get a confusing failure.

The existing `GET /api/quota` example just above (lines 132-135) uses double quotes and a literal placeholder (`"X-Quota-Management-Secret: my_secret"`), which is one safe pattern. Using double quotes around the variable form is another.

Suggested fix — pick one of:

```bash
# match the GET example style (literal placeholder)
    -H "X-Quota-Management-Secret: my_secret" \

# or, if keeping the env-var form, switch the outer quotes
    -H "X-Quota-Management-Secret: $REANA_QUOTA_MANAGEMENT_SECRET" \
```

michaelbuchar: Fixed!

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +164,70 @@ The response contains the new limit and the current usage for the given user and
...
+
+```console
+$ curl -X PATCH https://reana.example.org/api/quota \
+    -H 'Content-Type: application/json' \
+    -H "X-Quota-Management-Secret: $REANA_QUOTA_MANAGEMENT_SECRET" \
+    --data '{
+      "email": "john.doe@example.org",
+      "resource_type": "cpu",
+      "quota_period_months": 2,
+      "quota_period_start_at": "2026-05-01T00:00:00"
```
tiborsimko: 🤖 **Add a `Z` timezone designator to the example timestamp (NIT)**

The PATCH request body example uses a naive `"2026-05-01T00:00:00"`. The OpenAPI spec declares `quota_period_start_at` as `format: date-time` (RFC 3339), which requires a timezone designator, and the corresponding response-side serialisation in reana-server now appends `Z`. The marshmallow parser on this PATCH input still accepts a naive string, so this works in practice — but a user copy-pasting the example as a template might also use the same naive shape when reading the response back, which would fail strict RFC 3339 validation in the Python (bravado) client.

Suggested fix — align the example with the canonical form used elsewhere in the periodic-quota docs and specs:

```json
"quota_period_start_at": "2026-05-01T00:00:00Z"
```

michaelbuchar: Good catch thanks, fixed!

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +164,70 @@ The response contains the new limit and the current usage for the given user and
...
+manually via the administrative quota refresh command.
+
+To force an immediate refresh, run:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- reana-db quota resource-usage-update
+```
+
+This command advances the current CPU quota window if needed and recalculates
+the user's CPU usage for the current active period only. This is particularly
```
tiborsimko: 🤖 **"the user's CPU usage" misleads about the command's scope (HIGH)**

`reana-db quota resource-usage-update` processes *all users with active tokens* (see `update_users_cpu_quota` in `reana_db/utils.py` — the user-loop is unconditional unless an explicit user is passed, which the CLI doesn't do). The current wording ("recalculates **the user's** CPU usage for the current active period only") reads as if the command operates on a single user, which an admin might infer means there is a per-user variant or flag. There isn't.

Suggested fix:

```markdown
This command advances the current CPU quota window if needed and recalculates
each user's CPU usage for their current active period only. This is
particularly useful after changing quota period settings, when you want the
web UI and REST API responses to reflect the new accounting window
immediately.
```

(Also smoothes the "after changing... and wanting" parallel into a temporal clause.)

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +164,70 @@ The response contains the new limit and the current usage for the given user and
...
+```
+
+If you omit `--quota-period-start-at`, REANA preserves the currently stored
+period start. If no period start has been stored yet, REANA initialises it from
+the user account creation time and the chosen accounting cadence.
+
+You can also force-start a new CPU quota window for a user:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-replenish -e john.doe@example.org --quota-period-start-at 2026-07-01T00:00:00Z --admin-access-token $REANA_ACCESS_TOKEN
```
tiborsimko: 🤖 **`quota-replenish` is introduced without explaining how it differs from `quota-set-period --quota-period-start-at` (HIGH)**

Both commands accept `--quota-period-start-at` and both produce a new period start, so an admin reading the page side-by-side can't tell when to use one over the other. The Click help for `quota-replenish` describes it as *"Force start a new periodic quota window"* — which implies extra semantics beyond setting metadata (presumably it also resets accumulated usage, hence "replenish"). The docs should make that distinction explicit, otherwise the example reads as a redundant alternative to `quota-set-period`.

Suggested fix — add a one-sentence framing before the example:

```markdown
While `quota-set-period` updates the period metadata without touching
accumulated usage, you can also force-start a brand new CPU quota window
for a user (replenishing their quota — accumulated usage in the new
window starts at zero):
```

If the actual semantic differs from this guess, please correct accordingly — the key point is that the page should answer "why would I pick this over the previous command".

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +164,70 @@ The response contains the new limit and the current usage for the given user and
...
+
+## Refreshing quota usage after changing quota periods
+
+Changing `quota_period_months` or `quota_period_start_at` updates the stored
+period metadata immediately, but it does not by itself recompute the already
+accounted CPU usage for the new window.
+
+CPU usage is recalculated when the next quota usage update runs, either
+automatically via the configured `quota.periodic_update_policy` cronjob, or
+manually via the administrative quota refresh command.
```
tiborsimko: 🤖 **"the administrative quota refresh command" is a phantom name (LOW)**

The actual command three lines below is `reana-db quota resource-usage-update`. The phrase "the administrative quota refresh command" doesn't appear anywhere in the codebase or other docs, so a user grepping the docs for "refresh" finds nothing actionable, and the abstract label has to be re-resolved on every read.

Suggested fix — name the command directly:

```markdown
CPU usage is recalculated when the next quota usage update runs, either
automatically via the configured `quota.periodic_update_policy` cronjob, or
manually via the `reana-db quota resource-usage-update` command shown
below.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,49 @@ Individual users will be able to run as many workflows as they want, provided th
...
+As with default quota limits, changing the deployment-wide periodic CPU
+setting only affects newly created users. Existing users keep their
+current settings until they are updated manually.
+
+If you want to backfill the deployment default periodic CPU policy for already
+existing users, run:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- reana-db quota init-default-period
+Initialised periodic CPU reset policy for 42 existing users.
```
tiborsimko: 🤖 **Terminology drift across Helm value, DB columns, prose, and CLI output (LOW)**

The same concept is referred to as:

- `cpu_period_reset_months` (Helm value — uses "reset")
- `quota_period_months` / `quota_period_start_at` (DB columns + REST API fields — use "period")
- "periodic CPU accounting policy" (this page's prose)
- "periodic CPU reset policy" (the CLI success message printed here on line 87)
- "CPU accounting window" / "quota period" (elsewhere)

The word "reset" appears in the Helm key and the CLI output but is never explained in the concept prose — an operator may reasonably ask whether "reset" implies usage is wiped at period boundaries (it isn't; the period just rolls over and usage is recomputed inside the new window).

Suggested fix — pick one canonical pair (e.g. *CPU accounting window* + *quota period*) and align the CLI output line and the Helm value description to it. Alternatively, add a one-sentence note acknowledging that `cpu_period_reset_months` and `quota_period_months` are the same setting at two different levels (Helm-level default vs per-user value).

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +164,70 @@ The response contains the new limit and the current usage for the given user and
...
 }
 ```
+
+## Setting individual quota periods
+
+You can also configure user periodic CPU accounting settings. For example,
+to set a two-month accounting window for one user, run:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set-period -e john.doe@example.org --resource cpu --quota-period-months 2 --admin-access-token $REANA_ACCESS_TOKEN
```
tiborsimko, line 175: 🤖 **CLI shorthand inconsistency: `-r` vs `--resource` (LOW)**

Line 109 above uses `quota-set ... -r disk` (short form), while the new `quota-set-period` examples here use `--resource cpu` (long form). Both forms work for both commands, but the inconsistency makes the page slightly harder to skim. Pick one style — given the long lines elsewhere in this page, the long form is fine if used consistently everywhere; alternatively, switch the new examples to `-r cpu`.

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,49 @@ Individual users will be able to run as many workflows as they want, provided th
 !!! warning
     If the default quota values are changed, they will only be applied to newly created users. Users that already exist will not be affected by this change.
 
+## Setting default periodic CPU accounting windows
```
tiborsimko: 🤖 **Plural "windows" doesn't match the single knob being documented (LOW)**

The section discusses one configuration value (`cpu_period_reset_months`) that defines the length of *one* accounting window at a time. The plural reads slightly off.

Suggested fix:

```markdown
## Setting the default periodic CPU accounting window
```

docs/advanced-usage/user-quotas/index.md
```diff
@@ -55,6 +55,22 @@ corner of the web interface:
     means that the cluster administrators did not enable the user quotas feature on
     this REANA instance.
 
+## Understand periodic CPU quota windows
```
tiborsimko: 🤖 **Heading style inconsistent with sibling headings (LOW)**

Other H2 headings on this page are noun phrases ("Check quota limits and usage", "What to do when quotas are exceeded"). "Understand periodic CPU quota windows" is an imperative, which stands out.

Suggested fix — match the prevailing style:

```markdown
## Periodic CPU quota windows
```

or, if you'd like to keep the gerund hint:

```markdown
## Understanding periodic CPU quota windows
```

docs/advanced-usage/user-quotas/index.md
```diff
@@ -55,6 +55,22 @@ corner of the web interface:
...
+
+Some REANA deployments account CPU usage inside recurring quota periods, for
+example every three months, instead of over the whole lifetime of your account.
+Periodic accounting currently applies only to CPU quotas.
+
+When periodic CPU accounting is enabled, the profile page can show the current
```
tiborsimko: 🤖 **"can show" hedges, and the section is missing the customary profile-page screenshot (LOW)**

Two small style issues for the same section:

1. "the profile page **can show** the current CPU accounting window" — hedging. Either it shows the window (when the feature is on) or it doesn't. "can" doesn't add information.

2. Every other major section on this page that refers to the UI includes a screenshot (`ui-profile-your-quota.png`, `ui-sort-by-most-disk-quota-used.png`, `ui-delete-workflow.png`, etc.). Since reana-ui PR #494 adds visible `startLabel` / `endLabel` text to the profile-page quota box, a screenshot here would match the page's convention and help users recognise the feature at a glance.

Suggested fix:

```markdown
When periodic CPU accounting is enabled, the profile page shows the current
CPU accounting window together with your CPU usage:

![CPU accounting window on the profile page](/images/ui-profile-cpu-quota-period.png)
```

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Removed quota-replenish altogether

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed CLI wording

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: I’d keep this, both forms work, and I don’t think this needs a docs change.

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
```diff
@@ -55,6 +55,24 @@ corner of the web interface:
     means that the cluster administrators did not enable the user quotas feature on
     this REANA instance.
 
+## Periodic CPU quota windows
```
tiborsimko: 🤖 **Section title hides what the feature does**

"Periodic CPU quota windows" describes the mechanism but not the user-visible benefit (the quota renews every N months based on cluster admin settings). Reframing around "renewable" makes the value clearer to end users.

```suggestion
## Renewable CPU quotas
```

tiborsimko (commented):
🤖 Verbiage pass over the new periodic CPU quota docs.

The text mixes several overlapping nouns for the same concept — "quota window", "accounting window", "quota period", "active window", "accounting cadence" — and the user-facing section title hides the actual benefit (the quota resets/renews). Below are inline suggestions standardizing on **"CPU quota period"** (matching the `quota_period_months` / `quota_period_start_at` field names users will see in the API) and replacing the awkward verb "to account CPU time" with "to track CPU usage". Feel free to accept the suggestions selectively.

docs/advanced-usage/user-quotas/index.md
```diff
@@ -55,6 +55,24 @@ corner of the web interface:
...
+
+Some REANA deployments account CPU usage inside recurring quota periods, for
+example every three months, instead of over the whole lifetime of your account.
+Periodic accounting currently applies only to CPU quotas.
```
tiborsimko: 🤖 **Awkward verb "to account" and inconsistent noun**

"Account CPU usage" reads like a translation; "track CPU usage" is more natural English. Also dropping "periodic accounting" in favour of the user-facing framing.

```suggestion
Some REANA deployments track CPU usage inside recurring quota periods, for
example every three months, instead of over the whole lifetime of your account.
This renewable-quota mechanism currently applies only to CPU quotas.
```

docs/advanced-usage/user-quotas/index.md
```diff
@@ -55,6 +55,24 @@ corner of the web interface:
...
+
+When periodic CPU accounting is enabled, the profile page shows the current
+CPU accounting window together with your CPU usage. The start of each window is
+derived from your account creation time and from the quota period length chosen
+by the administrators.
```
tiborsimko: 🤖 **Standardize on "CPU quota period"**

Replaces "periodic CPU accounting", "CPU accounting window", and "quota period length" with the single noun "CPU quota period", matching the `quota_period_months` field name.

```suggestion
When renewable CPU quotas are enabled, the profile page shows the current
CPU quota period together with your CPU usage. The start of each period is
derived from your account creation time and from the period length chosen
by the administrators.
```

docs/advanced-usage/user-quotas/index.md
```diff
@@ -55,6 +55,24 @@ corner of the web interface:
...
+
+!["CPU accounting window on the profile page"](/images/ui-profile-cpu-quota-period.png)
```
tiborsimko: 🤖 **Alt text should match the chosen noun**

```suggestion
!["CPU quota period on the profile page"](/images/ui-profile-cpu-quota-period.png)
```

docs/advanced-usage/user-quotas/index.md
```diff
@@ -55,6 +55,24 @@ corner of the web interface:
...
+
+If the administrators change your CPU quota period, the new period start is
+stored immediately. However, the displayed CPU usage is recalculated only during
+the next quota usage update cycle. Until then, the usage number may still
+reflect the previous accounting window.
```
tiborsimko: 🤖 **"Previous accounting window" → "previous period"**

Keeps consistent terminology with the rest of the section.

```suggestion
If the administrators change your CPU quota period, the new period start is
stored immediately. However, the displayed CPU usage is recalculated only during
the next quota usage update cycle. Until then, the usage number may still
reflect the previous period.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
 !!! warning
     If the default quota values are changed, they will only be applied to newly created users. Users that already exist will not be affected by this change.
 
+## Setting default periodic CPU accounting window
```
tiborsimko: 🤖 **Wordy section title**

Four qualifiers ("default periodic CPU accounting window") for what is essentially the default CPU quota period.

```suggestion
## Setting the default CPU quota period
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+
+In addition to setting CPU usage limits, you can also configure REANA to
+account CPU time inside recurring quota periods instead of over the whole
+lifetime of a user account. Periodic accounting is currently supported only for
+the `cpu` resource.
```
tiborsimko: 🤖 **"account CPU time" → "track CPU time" + consistent noun**

"Account" as a transitive verb is unusual in English; "track" is the natural fit. Also unifying "periodic accounting" / "recurring quota periods" under one phrase.

```suggestion
In addition to setting CPU usage limits, you can also configure REANA to
track CPU time inside recurring quota periods instead of over the whole
lifetime of a user account. Renewable quota periods are currently supported
only for the `cpu` resource.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+
+To enable a deployment-wide default periodic CPU accounting policy for newly
+created users, set the following Helm value to a positive integer:
```
tiborsimko: 🤖 **Tighten "deployment-wide default periodic CPU accounting policy"**

Five nouns where two will do.

```suggestion
To enable a deployment-wide default CPU quota period for newly created
users, set the following Helm value to a positive integer:
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+
+```yaml
+quota:
+  cpu_period_reset_months: 3
+```
+
+Set `quota.cpu_period_reset_months` to `0` to disable the
+deployment-wide default periodic CPU accounting policy. This Helm-level
+default is stored per user as `quota_period_months`.
```
tiborsimko: 🤖 **Slightly redundant with the previous paragraph**

The "set to a positive integer" wording above already implies `0` disables. Trimming, and dropping "Helm-level default" jargon.

```suggestion
Set `quota.cpu_period_reset_months` to `0` to disable the deployment-wide
default. The chosen value is stored per user as `quota_period_months`.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+
+When this value is set, each newly created user receives:
+
+- `quota_period_months` which defines the length of the CPU accounting window
+- `quota_period_start_at` which defines the start of the current active window
```
tiborsimko: 🤖 **"CPU accounting window" / "current active window" → "CPU quota period" / "current period"**

Matches the noun used by the user-facing docs and by the API field names.

```suggestion
- `quota_period_months` which defines the length of the CPU quota period
- `quota_period_start_at` which defines the start of the current period
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+
+The current active window is derived from the user account creation time. For
+example, for a user created on `2026-04-14T13:06:32` and a three-month period,
+REANA will account CPU usage in windows such as `2026-04-14 -> 2026-07-14`,
+`2026-07-14 -> 2026-10-14`, and so on.
```
tiborsimko: 🤖 **Same noun standardization + `->` is unusual notation here**

Replacing the arrow with prose date ranges, which is the style used elsewhere in these docs.

```suggestion
The current period is derived from the user account creation time. For
example, for a user created on `2026-04-14T13:06:32` and a three-month period,
REANA will track CPU usage in successive periods such as `2026-04-14` to
`2026-07-14`, `2026-07-14` to `2026-10-14`, and so on.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+
+As with default quota limits, changing the deployment-wide periodic CPU
+setting only affects newly created users. Existing users keep their
+current settings until they are updated manually.
```
tiborsimko: 🤖 **"deployment-wide periodic CPU setting" → "deployment-wide CPU quota period"**

Clearer and consistent.

```suggestion
As with default quota limits, changing the deployment-wide CPU quota period
only affects newly created users. Existing users keep their current
settings until they are updated manually.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+
+If you want to backfill the deployment default CPU quota period for already
+existing users, run:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- reana-db quota init-default-period
+Initialised the default CPU quota period for 42 existing users.
```
tiborsimko, line 87: 🤖 **Check spelling against actual CLI output**

The rest of the REANA docs use US English ("initialize", "recalculate"); "Initialised" here stands out. This is literal CLI output, though, so the spelling should match whatever `reana-db quota init-default-period` actually prints — worth a quick check on the source side rather than a doc-only edit.

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -47,6 +47,50 @@ Individual users will be able to run as many workflows as they want, provided th
...
+```
+
+This command only initializes users who do not already have any custom periodic
+CPU settings.
```
tiborsimko: 🤖 **"any custom periodic CPU settings" → "a custom CPU quota period"**

Matches the noun used everywhere else.

```suggestion
This command only initializes users who do not already have a custom CPU
quota period configured.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
   "usage": 12000000
 }
 ```
+
+## Setting individual quota periods
```
tiborsimko: 🤖 **Clarify the section is CPU-only and per-user**

The page header already says "individual quota limits" above; this section is specifically about per-user CPU quota periods.

```suggestion
## Setting per-user CPU quota periods
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+
+You can also configure user periodic CPU accounting settings. For example,
+to set a two-month accounting window for one user, run:
```
tiborsimko: 🤖 **"periodic CPU accounting settings" / "accounting window" → consistent noun**

```suggestion
You can also configure CPU quota periods on a per-user basis. For example,
to set a two-month CPU quota period for one user, run:
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set-period -e john.doe@example.org --resource cpu --quota-period-months 2 --admin-access-token $REANA_ACCESS_TOKEN
+Periodic quota fields updated successfully.
+```
+
+You can optionally provide the current active window start explicitly:
```
tiborsimko: 🤖 **"current active window" → "current period"**

```suggestion
You can optionally provide the current period start explicitly:
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set-period -e john.doe@example.org --resource cpu --quota-period-months 2 --quota-period-start-at 2026-05-01T00:00:00Z --admin-access-token $REANA_ACCESS_TOKEN
+Periodic quota fields updated successfully.
+```
+
+If you omit `--quota-period-start-at`, REANA preserves the currently stored
+period start. If no period start has been stored yet, REANA initialises it from
+the user account creation time and the chosen accounting cadence.
+The `--quota-period-start-at` value may also be set in the future.
+Until that date is reached, CPU usage is not counted in the new periodic
+window yet, so the user effectively has unlimited quota available.
```
tiborsimko: 🤖 **Two issues here**

1. "the chosen accounting cadence" introduces yet another term for the same thing — replace with "the configured period length".
2. "the user effectively has unlimited quota available" is not quite right: the limit still applies, but nothing is being recorded against it yet. Reframing avoids the misleading suggestion that the quota itself is unlimited.

```suggestion
If you omit `--quota-period-start-at`, REANA preserves the currently stored
period start. If no period start has been stored yet, REANA initialises it from
the user account creation time and the configured period length.
The `--quota-period-start-at` value may also be set in the future.
Until that date is reached, no CPU usage is recorded against the new period
yet.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+
+To force-start a new CPU quota window for a user, reuse
+`quota-set-period` and provide only the new `--quota-period-start-at`:
```
tiborsimko: 🤖 **"CPU quota window" → "CPU quota period"**

```suggestion
To force-start a new CPU quota period for a user, reuse
`quota-set-period` and provide only the new `--quota-period-start-at`:
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+      "resource_type": "cpu",
+      "quota_period_months": 2,
+      "quota_period_start_at": "2026-05-01T00:00:00Z"
+    }'
+```
+
+The REST API accepts one or both of `quota_period_months` and
+`quota_period_start_at`.
+
+## Refreshing quota usage after changing quota periods
```
tiborsimko: 🤖 **"Refreshing" suggests a UI action; "recalculating" is what actually happens**

Also the section deals with one period at a time, so singular reads better.

```suggestion
## Recalculating CPU usage after changing a quota period
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+
+Changing `quota_period_months` or `quota_period_start_at` updates the stored
+period metadata immediately, but it does not by itself recompute the already
+accounted CPU usage for the new window.
```
tiborsimko: 🤖 **"already accounted CPU usage" → "accumulated CPU usage" + consistent noun**

```suggestion
Changing `quota_period_months` or `quota_period_start_at` updates the stored
period metadata immediately, but it does not by itself recompute the
accumulated CPU usage for the new period.
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- reana-db quota resource-usage-update
+```
+
+This command advances the current CPU quota window if needed and recalculates
+each user's CPU usage for their current active period only. This is
+particularly useful after changing quota period settings, when you want the
+web UI and REST API responses to reflect the new accounting window
+immediately.
```
tiborsimko: 🤖 **Final consistency pass: "window" / "active period" / "accounting window" → "period"**

```suggestion
This command advances the current CPU quota period if needed and recalculates
each user's CPU usage for the current period only. This is particularly
useful after changing quota period settings, when you want the web UI and
REST API responses to reflect the new period immediately.
```

tiborsimko (commented):
🤖 Follow-up: missing "As of REANA X.Y" version-introduction phrases on the new sections.

The REANA docs use this pattern to mark when a feature became available within a longer-lived page. Examples already in the docs:

- `As of REANA 0.95 release series, you can also set quota limits for specific users using the Quota management REST API...` — `configuring-user-quotas/index.md:118`
- `As of version 0.9, REANA administrators can configure a global workspace file retention period...` — `configuring-global-workspace-retention-rules/index.md:3`
- `As of version 0.8, REANA supports two different workflow scheduling strategies:` — `configuring-scheduler/index.md:5`

None of the four new sections in this PR carry such a notice. The current chart version is `0.95.0-alpha.4`, so the feature will land in the **REANA 0.95 release series**. Below are three inline suggestions that weave the version notice into the opening sentence of each new feature section (the fourth, `## Refreshing quota usage…`, is an operational follow-on to the per-user section and doesn't need its own marker).

Note: these three suggestions target the same lines as the body-paragraph rewrites in my earlier review and fold the version notice in — accept these in place of the earlier ones for those paragraphs.

docs/administration/configuration/configuring-user-quotas/index.md
tiborsimko: 🤖 **Add "As of REANA 0.95 release series" version notice (supersedes earlier body rewrite for this paragraph)**

Folds the version notice into the opening sentence, in the same style as line 118 below.

```suggestion
As of REANA 0.95 release series, in addition to setting CPU usage limits, you
can configure REANA to track CPU time inside recurring quota periods instead
of over the whole lifetime of a user account. Renewable quota periods are
currently supported only for the `cpu` resource.
```

docs/advanced-usage/user-quotas/index.md
tiborsimko: 🤖 **Add "As of REANA 0.95 release series" version notice (supersedes earlier body rewrite for this paragraph)**

Mirrors the `Since REANA 0.8.0` notice at the top of the page, which marks the quota feature as a whole — this new sub-feature deserves its own marker.

```suggestion
As of REANA 0.95 release series, some deployments track CPU usage inside
recurring quota periods, for example every three months, instead of over the
whole lifetime of your account. This renewable-quota mechanism currently
applies only to CPU quotas.
```

docs/administration/configuration/configuring-user-quotas/index.md
tiborsimko: 🤖 **Add "As of REANA 0.95 release series" version notice (supersedes earlier body rewrite for this paragraph)**

This section is far enough below the earlier renewable-quota section that a reader landing here directly (e.g. via search) benefits from its own version marker.

```suggestion
As of REANA 0.95 release series, you can also configure CPU quota periods on
a per-user basis. For example, to set a two-month CPU quota period for one
user, run:
```

docs/administration/configuration/configuring-user-quotas/index.md
```diff
@@ -121,3 +165,76 @@ The response contains the new limit and the current usage for the given user and
...
+CPU usage is recalculated when the next quota usage update runs, either
+automatically via the configured `quota.periodic_update_policy` cronjob, or
+manually via the `reana-db quota resource-usage-update` command shown
+below.
+
+To force an immediate refresh, run:
+
+```console
+$ kubectl exec -i -t deployment/reana-server -- reana-db quota resource-usage-update
+```
```
tiborsimko, line 235: 🤖 **Trigger the existing cronjob instead of `kubectl exec`-ing the server pod**

See the top-level comment for the rationale. The job-name suffix `$(date +%s)` avoids collisions if an operator needs to trigger the refresh more than once.

````suggestion
CPU usage is recalculated when the next quota usage update runs, either
automatically on the configured `quota.periodic_update_policy` schedule, or
manually by triggering an ad-hoc run of the same cronjob.

To force an immediate refresh, spawn a one-off Job from the existing
`reana-resource-quota-update` cronjob:

```console
$ kubectl create job --from=cronjob/reana-resource-quota-update \
    reana-resource-quota-update-manual-$(date +%s)
```
````

**Review by tiborsimko — COMMENTED** [review]
🤖 Follow-up: advertise triggering the existing cronjob rather than running the underlying command directly.

The `reana-resource-quota-update` cronjob created by the Helm chart literally runs `reana-db quota resource-usage-update` (see `reana/helm/reana/templates/cronjobs.yaml:156`), so the two paths produce identical results. But spawning a one-off Job from the cronjob is preferable for several reasons:

- **Inherits the cronjob's full spec** — same image, same env vars (`REANA_PERIODIC_RESOURCE_QUOTA_UPDATE_POLICY=true`), same secrets, node selector, etc. — so a manual run always matches the periodic one byte-for-byte.
- **Respects maintenance mode** — the cronjob carries `suspend: true` when `maintenance.enabled` is set; `kubectl exec` into the server deployment bypasses that gate.
- **Proper Job lifecycle** — a one-off Job produces its own pod, logs and status, and gets garbage-collected via `failedJobsHistoryLimit`, instead of running inside the long-lived `reana-server` pod where logs are interleaved with everything else.
- **Decouples the doc from internal command structure** — if the subcommand is later renamed or gains flags, the chart updates the cronjob and operators following these docs don't have to update their procedure.

The standard kubectl recipe is `kubectl create job --from=cronjob/<name> <job-name>`. Suggested rewording inline.

File: docs/advanced-usage/user-quotas/index.md (hunk 14 above)
Comment (tiborsimko) [inline docs/advanced-usage/user-quotas/index.md]: 🤖 **Apply the `screenshot-browser-mockup` CSS class (supersedes earlier alt-text suggestion for this line)**

Combines the alt-text rename from the first review with the macOS-chrome class.

```suggestion
!["CPU quota period on the profile page"](/images/ui-profile-cpu-quota-period.png){.screenshot-browser-mockup}
```

tiborsimko (commented):
🤖 Follow-up: the new screenshot is missing the macOS-style browser frame that the other images on the page have.

The other screenshots on this page (`ui-profile-your-quota.png`, `ui-delete-workflow.png`, `ui-sort-by-most-disk-quota-used.png` …) have the macOS window chrome — dark titlebar with red/yellow/green traffic-light buttons — baked into the PNG, because they were captured natively on macOS via `Cmd+Shift+4 → Space → click window`. The new `ui-profile-cpu-quota-period.png` was captured without that chrome and stands out visually.

The docs already ship a helper for exactly this situation. Commit `3659eb3` (Aug 2023) introduced an `img.screenshot-browser-mockup` CSS class in `docs/stylesheets/extra.css:178` that composites a synthetic titlebar with traffic-light buttons (background image `docs/images/window-buttons-for-mockup.png`) on top of any image tagged with `{.screenshot-browser-mockup}`. The `attr_list` markdown extension is already enabled in `mkdocs.yml` to support the syntax. The helper is currently used by 6 images, e.g. `docs/running-notebooks/opening-sessions/index.md:19` and `docs/administration/configuration/configuring-gitlab-integration/index.md:23`.

Two options to fix the inconsistency:

- **Option A (one-line markdown change):** append `{.screenshot-browser-mockup}` to the image reference and let the CSS render the chrome. Cheapest fix, matches the established docs convention. The result won't be pixel-identical to the macOS-native captures (synthetic titlebar is 20px tall vs ~40px, no tab/`+` button, no rendered tab label), but it carries the same visual signal.
- **Option B (higher fidelity):** re-capture the screenshot on macOS with `Cmd+Shift+4 → Space → click window` so the native chrome is baked into the PNG, matching `ui-profile-your-quota.png` exactly.

Inline suggestion below for Option A — folds together with my earlier alt-text suggestion for this line.

docs/advanced-usage/user-quotas/index.md
tiborsimko: Alternatively, just crop the screenshot to show only the most important part, i.e. the quota period. It is not necessary to reproduce the whole browser window again. This will better attract eyes towards what's most important WRT quota periods.

Screenshot 2026-05-18 at 15 02 43

BTW in the above image the limit is too generous, and the percentage is not very instructive (0%). It'll be better to retake the screenshot, set a limit to say 15 minutes, so that a consumption of about 2 minutes would create 13% of some such.

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Kept as is since it does return "Initialised"

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/administration/configuration/configuring-user-quotas/index.md
michaelbuchar: Fixed

docs/advanced-usage/user-quotas/index.md
michaelbuchar: Fixed, and updated the screenshot with a more informative limit!

tiborsimko (approved):

PR #225, opened 2026-05-18
docs(administration): document token revocation management

michaelbuchar:
Closes reanahub/reana-server#776

### DIFF
--- a/docs/administration/management/managing-users/index.md
+++ b/docs/administration/management/managing-users/index.md
@@ -47,6 +47,40 @@ $ kubectl exec -i -t deployment/reana-server -- flask reana-admin token-revoke -
 User token c0fa47fa00ae4013a13fd7n (new.web.user@example.org) was successfully revoked.
 ```
 
+If you need to revoke user tokens remotely without Kubernetes cluster access,
+you can optionally enable a token-management REST API endpoint:
+
+```yaml
+components:
+  reana_server:
+    environment:
+      REANA_TOKEN_MANAGEMENT_SECRET: "my_secret"
+```
+
+When this secret is non-empty, you can revoke a user's active token with:
+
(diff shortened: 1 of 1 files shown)


docs/administration/management/managing-users/index.md
```diff
@@ -47,6 +47,31 @@ $ kubectl exec -i -t deployment/reana-server -- flask reana-admin token-revoke -
...
+      REANA_TOKEN_MANAGEMENT_SECRET: "my_secret"
+```
+
+When this secret is non-empty, you can revoke a user's active token with:
+
+```console
+$ curl -X DELETE https://reana.example.org/api/token \
+    -H 'Content-Type: application/json' \
+    -H "X-Token-Management-Secret: $REANA_TOKEN_MANAGEMENT_SECRET" \
+    --data '{"email": "new.web.user@example.org"}'
```
tiborsimko, line 66: 🤖 **Mention that the endpoint also accepts `user_id`**

The endpoint accepts either `email` or `user_id` to identify the target user (exactly one of the two), but the docs only show the `email` form. A reader who's automating this against a system where emails aren't unique or stable may not realise they have a UUID option. One-line addition right after the cURL block is enough:

```diff
     --data '{"email": "new.web.user@example.org"}'
 ```

+You can also identify the user by REANA user ID instead of email, e.g.
+`--data '{"user_id": "aa37d63d-3186-45d5-aa40-5d221cb170bf"}'`.
+
 If `REANA_TOKEN_MANAGEMENT_SECRET` is empty, this endpoint is disabled.
```

michaelbuchar: Fixed

tiborsimko (approved):

PR #228, opened 2026-07-03
docs(slurm): update default partition to photon

tiborsimko:
Align the documented `slurm_partition` default to `photon` after the CERN Slurm `inf-short` partition was retired.

Use `phodev` in the explicit partition example so the workflow snippet continues to demonstrate an override of the default value.

Closes reanahub/reana-job-controller#529


michaelbuchar:
Small documentation consistency point: this issue series now introduces SLURM_HOSTNAME and SLURM_GSS_HOSTNAME as supported configuration knobs for the Slurm backend, but the admin guide still says that the Slurm integration cannot be customised via Helm values or environment variables. Maybe we can also update docs/administration/configuration/configuring-compute-backends/index.md as part of this work as well? Otherwise operators following the admin docs may miss the new workaround.

tiborsimko:
> Small documentation consistency point: this issue series now introduces SLURM_HOSTNAME and SLURM_GSS_HOSTNAME as supported configuration knobs for the Slurm backend, but the admin guide still says that the Slurm integration cannot be customised via Helm values or environment variables. Maybe we can also update docs/administration/configuration/configuring-compute-backends/index.md as part of this work as well? Otherwise operators following the admin docs may miss the new workaround.

Good catch, thanks! The categorical statement was stale. I’ve updated the admin guide to point to the full list in the Helm chart documentation, described the `SLURM_HOSTNAME` and `SLURM_GSS_HOSTNAME` settings, and extended the configuration example accordingly.

docs/administration/configuration/configuring-compute-backends/index.md
```diff
@@ -152,9 +152,13 @@ and customise the
...
 `reanahub/reana-job-controller-htcondorcern-slurmcern:0.9.0`.
 
-Currently it is not possible to further customise the Slurm integration through
-additional Helm values or environment variables. Please get in touch should you
-need to customise the Slurm integration further in your REANA deployment.
+The Slurm integration can be further customised by setting environment
+variables under `components.reana_job_controller.environment`; see the
+[REANA Helm chart documentation](https://github.com/reanahub/reana/tree/master/helm/reana)
+for the full list. For example, `SLURM_HOSTNAME` selects the Slurm head node,
+and `SLURM_GSS_HOSTNAME` explicitly selects the Kerberos service principal
```
michaelbuchar: Nice improvement to the Slurm configuration guidance. One consistency issue: this section introduces SLURM_GSS_HOSTNAME, but the deployment example still uses reana-job-controller-htcondorcern-slurmcern:0.9.0. That release predates this environment variable and also retains the old inf-short default, so operators copying the example would not receive the behaviour documented here.
Could we replace the hard-coded tag with a release containing reana-job-controller #528, or use a clearly marked <version> placeholder with a note that the selected version must include the new Slurm configuration support?

tiborsimko: Since the 0.95.0 image has not been published yet, I kept the example on the real, pullable `0.9.0` tag. The example configures only `SLURM_HOSTNAME`, which is already supported by that release; it does not use `SLURM_GSS_HOSTNAME`. And I clarified that `SLURM_GSS_HOSTNAME` is additionally available starting from the REANA 0.95.0 release series. This makes the version boundary explicit while keeping the example copy-pasteable. We can upgrade the snippets once REANA 0.95.0 is officially released.

michaelbuchar (approved):
LGTM!

